PR #89, opened 2015-12-03
Set default location to Penn, when no gps

### LINKED ISSUE #85: Center map on campus before location loads rather than 0,0
Coordinates for center of campus are roughly 39.9529, -75.197098

kbharathala:
@adelq 

Fixed #85


adelq:
LGTM

PR #113, opened 2016-03-07
Krishna more laundry fixes

kbharathala:
- fixed bug which crashed laundry when pressing back
- changed headerstyle to match everything else
- removed extraneous files

@dlakata


dlakata:
LGTM

PR #123, opened 2017-04-24
Study Room Booking, Dining, Laundry fixes

joshdoman:
All unused files have been removed, including the unfinished homepage.


adelq:
It's really difficult to review these massive code dumps, can you elaborate a bit on what this is exactly? All the commits have the message "removed unused files" but looks like there a ton of new files too, so would like to know what's actually in this pull request.

Is this the deployed version of the app with all of Krishna and Victor's changes?

Also, why did you change the permissions of a ton of files to executable?

joshdoman:
@adelq Sorry about the code dumps! I understand if it can't be reviewed. This is the current deployed version of the app with all changes to this date made. I deleted a ton of files that are no longer being used/will never be used again (such as the old dining page in obj-c) as well as a ton of files from the unfinished homepage. Once this gets merged into master, we can then open a new branch for the homepage, manually insert the files and use git properly. Can you post an example of a file changed to executable? I did not do so explicitly, but it may be something that the new version of XCode did on its own.

adelq:
Perfect, that's what I want to hear, thanks for the explanation. So moving forward, all development will be done on GitHub and merged into master, that's awesome.

If you go to the "File changed" tab, it indicates them with `100644 → 100755`. I wasn't able to check every file, but it seems that a large number of files were switched to executable in commit 30e4cd5ba2ce9935226cfa7e19fdc4a144c255f1. Ctrl-F of that text says 348 matches in this pull request. Some examples just to be specific:
- AGaramond.otf
- Assets/1920.png
- Assets/34-Blue.png

joshdoman:
@adelq Thank you that means a lot! You're absolutely right, this is the start of something new. Git forever and always!

PR #138, opened 2018-02-04
GSR Overhaul

joshdoman:
Merging the new GSR feature into development. The previous version's API relied on a previous version of the library booking site. The new features makes use from a new API.


zhileiz:
looks good to me~

PR #142, opened 2018-02-21
Dining + Laundry User Preferences

joshdoman:
Added network calls to server to track user preferences. Dining preferences are sent whenever the user clicks on a cell in the Dining Page, and laundry preferences are sent when the new version of the app first opens up and each time the user saves a new set of laundry rooms.


zhileiz:
LGTM~

PR #182, opened 2018-10-14
Merge for update 3.4.3

DominicHolmes:
Small UI updates & fixes, restoration of web view for dining details


DominicHolmes:
Hey everyone -- this is a merge request for development -> master, to publish update 3.4.3. We always build our App Store updates from the master branch.

Please comment LGTM if you approve!

sanieldalib:
For the most part LGTM. Only concern is that replacing of "Closed Today" text with empty string may be slightly confusing for users in those cases.

DominicHolmes:
@sanieldalib Yeah -- if there are no hours in the API call, it means its either closed or the API is down.

But before, we were saying "Closed Today" when the API goes down, when they don't have wifi, etc... Which is really bad (& an oversight on my part).

We can think about checking specific cases once the networking logic is refactored, but for now I want to get "Closed Today" out of the app.

benleim:
LGTM

PR #186, opened 2018-10-21
removed cacheing, hardcoded functions from dining

DominicHolmes:


DominicHolmes:
Dining data is now requested fresh, each time a user opens the page. Refreshing by pulling up on the slider actually refreshes data.

PR #190, opened 2018-11-03
Finished Status Bar and Networking Cleaned Up

sanieldalib:
Added a generic status bar that can be used anywhere with a navigation controller - currently only implemented on Dining Screen. Added callbacks to the getRequest() func to pass errors and status codes across controllers.


benleim:
LGTM my dood.

caringan:
LGTM

DominicHolmes:
@sanieldalib the status bar isn't disappearing correctly. If it appears on dining then you switch to GSR, you can see the red glow behind the translucent nav bar. make sure it disappears completely

PR #202, opened 2019-02-10
Settings for Dining Cell

sanieldalib:
Added option to select prefs for Dining Cell. If the user selects no options it defaults to the default 3 cafes. Otherwise, it shows the user's prefs. See photos:

img_0446
img_0447


PennMobile/General/Networking + Analytics/UserDBManager.swift
```diff
@@ -16,7 +16,7 @@ func getDeviceID() -> String {
 
 class UserDBManager: NSObject {
     static let shared = UserDBManager()
-    fileprivate let baseUrl = "https://api-dev.pennlabs.org"
+    fileprivate let baseUrl = "http://api.pennlabs.org"
```
joshdoman: Make this https

PennMobile/Dining/Networking/DiningAPI.swift
```diff
@@ -14,6 +14,7 @@ class DiningAPI: Requestable {
     static let instance = DiningAPI()
     
     let diningUrl = "https://api.pennlabs.org/dining/venues"
+    let diningPrefs =  "https://api-dev.pennlabs.org/dining/preferences/V2"
```
joshdoman: Are we still using this? I think this is happening in UserDBManager now

PennMobile/Home/Cells/Dining/HomeDiningCell.swift
```diff
@@ -212,15 +232,27 @@ extension HomeDiningCell {
...
     }
-    
+
+    fileprivate func getSettingsButton() -> UIButton {
+        let button = UIButton()
+        button.setTitleColor(.navigationBlue, for: .normal)
+        button.tintColor = .navigationBlue
+        button.setTitleColor(.secondaryTitleGrey, for: .highlighted)
+        button.setImage(#imageLiteral(resourceName: "settingsButton").withRenderingMode(.alwaysTemplate), for: .normal)
+        button.titleLabel?.font = .footerTransitionFont
```
joshdoman: I believe this can be deleted

PennMobile/Home/Cells/Dining/HomeDiningCell.swift
```diff
@@ -212,15 +232,27 @@ extension HomeDiningCell {
         label.translatesAutoresizingMaskIntoConstraints = false
         return label
     }
-    
+
+    fileprivate func getSettingsButton() -> UIButton {
+        let button = UIButton()
+        button.setTitleColor(.navigationBlue, for: .normal)
```
joshdoman: I believe this can be deleted

PennMobile/Home/Cells/Dining/HomeDiningCell.swift
```diff
@@ -212,15 +232,27 @@ extension HomeDiningCell {
...
+        button.tintColor = .navigationBlue
+        button.setTitleColor(.secondaryTitleGrey, for: .highlighted)
```
joshdoman: Same with this

PennMobile/Home/Controllers/DiningCellSettingsController.swift
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    func handleCancel() {
+        self.dismiss(animated: true, completion: nil)
+    }
+
+    func handleSave() {
+        var chosenVenues = chosenCafes.map {DiningVenue(venue: $0)}
+//        var chosenSet = Set<DiningVenue>.init()
```
joshdoman: Can we delete this comment? I don't think we need to convert to a set since duplicates aren't possible. Maintaining the array keeps the original ordering too

PennMobile/Home/Cells/Dining/HomeDiningCell.swift
sanieldalib: Yes the image is originally black

PennMobile/Home/Controllers/DiningCellSettingsController.swift
sanieldalib: Word -ill do this

PennMobile/Home/Cells/Dining/HomeDiningCell.swift
joshdoman: I might be wrong, but I think you can just set tintColor to make the image blue. I'm pretty sure setTitleColor doesn't do anything since the button doesn't have text. Can you try deleting setTitleColor but keeping tintColor and seeing what happens?

sanieldalib: whoops i read that wrong, will make that change

PR #206, opened 2019-02-17
Lots of Home Page Fixes + Features

sanieldalib:
Resolved a ton of smaller issues as well as a few larger features and fixes: 

Home Screen Bugs
- Added dining cell prefs
- Venues don't persist
- refresh Hangs
- Settings Buttons Changed
- Favorited Laundry appears on home
- dining open web view instead

More Tab
- Remove Fling and events
- Add Fitness (get design from grace)
- About at bottoms


PennMobile/Home/Controllers/DiningCellSettingsController.swift
```diff
@@ -0,0 +1,115 @@
...
+        }
+    }
+
+    func handleCancel() {
+        self.dismiss(animated: true, completion: nil)
+    }
+
+    func handleSave() {
+        var chosenVenues = chosenCafes.map {DiningVenue(venue: $0)}
+        print(chosenVenues.count)
```
joshdoman: @sanieldalib Make sure to remove all these print statements before merging.

sanieldalib: word - will do rn

PennMobile/Dining/Model/Venues/DiningVenueName.swift
```diff
@@ -10,29 +10,33 @@ import Foundation
...
-    case nch =           "NCH Dining"
+    case mcclelland =    "McClelland Express"
+    case nch =           "New College House"
     case hill =          "Hill House"
     case english =       "English House"
-    case falk =          "Falk Kosher"
-    case frontera =      "Tortas Frontera"
-    case gourmetGrocer = "Gourmet Grocer"
+    case falk =          "Falk Kosher Dining"
+    case frontera =      "Tortas Frontera at the ARCH"
```
joshdoman: What is the effect of this extra long name on the regular dining cell on the Dining Page? DiningCell.swift only uses the shortened name on the homepage.

PennMobile/Home/Controllers/HomeViewController.swift
```diff
@@ -164,7 +164,7 @@ extension HomeViewController {
 
     func fetchCellData(for itemTypes: [HomeCellItem.Type], _ completion: (() -> Void)? = nil) {
         let items = tableViewModel.getItems(for: itemTypes)
-        self.fetchCellData(for: items)
+        self.fetchCellData(for: items, completion)
```
joshdoman: Good catch!

joshdoman:
@sanieldalib Were you able to fix the laundry bug where it doesn't update the cell w/o restarting the app after selecting/deleting room(s)? Also, did we add code that refetches the entire model once every 30 min/1 hour/1 day?

PennMobile/Dining/Model/Venues/DiningVenueName.swift
sanieldalib: if I don't use the full name, the dictionary in storage (DiningVenue.swift init ) will not match it up, and the id won't get pushed to the server.

PennMobile/Home/Controllers/HomeViewController.swift
sanieldalib: Haha thanks, that's why the refresh wouldn't ever end.

PennMobile/Dining/Model/Venues/DiningVenueName.swift
joshdoman: I understand. I guess the question is do we need to use the short names in the Dining Page as well as in the Home Page because some are so much longer now. Alternatively, we could create a "Normal Names" enum/function which would handle special cases like Frontera where the name on the server is too long even for the full page. 

I think we're ok not worrying about this for now, but it's definitely something we should fix down the road, not to mention overhauling the venue identification system. It shouldn't break every time they change the names.

PennMobile/Dining/Networking/DiningAPI.swift
```diff
@@ -51,10 +51,6 @@ class DiningAPI: Requestable {
             completion(html)
         }
     }
-    
-    func saveDiningSettings(venues: [DiningVenueName], completion: (Bool) -> ()) {
-
```
joshdoman: Thanks for removing this too! I didn't see that.

PR #207, opened 2019-02-21
Home Loading Cell

sanieldalib:
Added loading icon to the home page that is displayed as data is loaded from the server initially


joshdoman:
No longer relevant as this has been improved and added manually.

PR #208, opened 2019-02-22
Migrated to Firebase Analytics

joshdoman:
Swapped out GoogleAnalytics for FirebaseAnalytics


PennMobile/General/Networking + Analytics/DatabaseManager.swift
```diff
@@ -35,7 +35,7 @@ enum DBError: String, LocalizedError {
     
     internal var sessionStarted = false
     
-    var dryRun: Bool = false //prevents any requests being sent
+    var dryRun: Bool = true //prevents any requests being sent
```
sanieldalib: Make sure we change this to false before release

PennMobile/General/Protocols.swift
```diff
@@ -51,7 +51,7 @@ protocol Trackable {}
 extension Trackable where Self: UIViewController {
     func trackScreen(_ name: String?) {
         if let name = name {
-            GoogleAnalyticsManager.shared.trackScreen(name)
+            FirebaseAnalyticsManager.shared.trackScreen(name)
```
sanieldalib: What is this supposed to do? The func is empty lol

PR #218, opened 2019-03-17
Home Page Reservations Cell

mgarciaferreiro:
Still need to fix constraints

image


joshdoman:
@mgarciaferreiro Did you make the height of the cell shorter? If so, can you post an image here?

mgarciaferreiro:
> @mgarciaferreiro Did you make the height of the cell shorter? If so, can you post an image here?

image

image

I did it to match the spacing of the dining cell and also changed the font to regular for uniformity, but I can change it back if it looks too tight

joshdoman:
Hmm yeah I'm not a huge fan. Personally I thought the original layout was beautiful, and I'd rather change how Dining looks on the home page than change how reservation looks. Also, it seems like you changed the layout for the reservation cell on BOTH the homepage and the reservation page. Is that what you intended?

mgarciaferreiro:
> Hmm yeah I'm not a huge fan. Personally I thought the original layout was beautiful, and I'd rather change how Dining looks on the home page than change how reservation looks. Also, it seems like you changed the layout for the reservation cell on BOTH the homepage and the reservation page. Is that what you intended?

I did, just because I wasn't sure how to change the height of the whole cell just for the homepage. I agree that the original layout looked nicer, so we should change dining for uniformity. Do you think it's better to just change dining in the homepage or also in the dining page? It might be off putting for them to be different

joshdoman:
Great glad we're on the same page. With regards to dining, I'd suggest that we not overwrite the original dining cell and instead create a copy of it that we change, like DiningCellV2 or something. Depending on how it looks, we may or may not want it to be only on the home page. One reason we may keep it only on the homepage is that the homepage uses

PR #228, opened 2019-05-01
Dining Balance

mgarciaferreiro:
Simulator Screen Shot - iPhone 6 - 2019-05-01 at 18 14 34


joshdoman:
@mgarciaferreiro If a user 2FA expires and we can't use their pennkey and password to get their dining dollars, do we pop up a view on the dining page asking the user to log in again?

mgarciaferreiro:
> @mgarciaferreiro If a user's 2FA expires and we can't use their pennkey and password to get their dining dollars, do we pop up a view on the dining page asking the user to log in again?

I don't think we currently do, but it might be a good idea to do it

PR #233, opened 2019-09-09
GSR UX Improvement

joshdoman:
Subtle but noticeable UX improvement for selecting GSR timeslots. Rather than not allowing the user to select an illegal timeslot (different room, different hours, etc.), we deselect the previous ones. If the user tries to deselect a middle timeslot in a 2 hour booking, we automatically deselect the timeslots to the right, keeping the selection legal.

Demo

### DIFF
--- a/PennMobile/GSR-Booking/ViewModel/GSRViewModel.swift
+++ b/PennMobile/GSR-Booking/ViewModel/GSRViewModel.swift
@@ -196,16 +196,72 @@ extension GSRViewModel: GSRSelectionDelegate {
         return true
     }
     
+    private func isValidAddition(timeSlot: GSRTimeSlot) -> Bool {
+        if currentSelection.isEmpty {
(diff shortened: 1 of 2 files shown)


PennMobile/GSR-Booking/ViewModel/GSRViewModel.swift
```diff
@@ -160,46 +160,32 @@ extension GSRViewModel: GSRSelectionDelegate {
...
-        } else if let prev = timeSlot.prev, let next = timeSlot.next,
-            currentSelection.contains(prev) && currentSelection.contains(next) {
-            return false
-        }
-        return true
+        
+        return isSameRoom && isBeforeOrAfter
     }
     
+    // Returns the number to the right removed if any
```
DominicHolmes: Please update this comment -- function doesn't return a number

PR #234, opened 2019-09-16
Require explicit refresh to update balance if log in fails

joshdoman:
Stop trying to log in to Campus Express to get dining balance if the previous attempt failed. Require the user to explicitly refresh their balance.


joshdoman:
Alternatively, we can pop up a login screen for the user if the attempt fails. Right now, the only way to refresh your login is to press refresh.

PR #250, opened 2019-10-22
Full Dining rewrite

DominicHolmes:
This PR includes a rewrite of our Dining-related classes using Codable.

These are some of the new files:
### DiningVenue
* `DiningVenue.swift` contains the basic structure of a DiningVenue, including its nested structs `MealsForDate` and `Meal`. This struct is all you need to understand what fields a DiningVenue has. When features are added going forward (such as menus), they should be represented in this file.
* `DiningVenue+Codable.swift` contains all parsing logic. This implements custom Encoding and Decoding flows for DiningVenue and its nested structs. We have _seperate flows_ for encoding and decoding to enable cacheing - I'll explain this more below.
* `DiningVenue+Extensions.swift` contains UI-related extensions (like .isOpen) and hour formatting logic
* `DiningAPIResponse.swift` contains the actual structure of the API response we receive

 ### Cache
* I've implemented a LocalJSONStore class that can encode any Codable struct in a single line. This uses the file system. When initializing the store, we can choose between permanent and cache data (the latter may be removed when the system is low on resources).
* The cache essentially creates a JSON file using the Encodable protocol, and stores that in the file system.
* When the app opens (after the first time) the cache loads much faster than an API request can be completed. The cache also persists when the user is offline. This is necessary, because we no longer have any hardcoded DiningVenues in the app.
* One slight hiccup - the cache is trying to decode the stored JSON in the same way the API response would be decoded. What this means is that we need to store data such that it _looks like_ the API structure. However, when I decode the API I parse some of the fields into a specific format, its not 1-to-1. Therefore, I implement custom `encode()` functions where necessary (on DiningVenue, for example) to make sure that when we cache JSON, we cache it to look like the server.

### Images
* I removed all images from the asset catalog. This should decrease our bundle size by around 3mb!
* Image loading is all done through kingfisher (which caches its own images). My cacheing system stores the URLs.
* Images now have a default grey background, as they won't always be loaded and ready.

### Testing
* I've done _some_ quality assurance, but more is definitely required.
* In particular, I made sure the cache is flexible and responds well to changing values (they won't be decoded, and then it will be overwritten after the next API response).
* I tried changing the date and time of my phone, while both online and offline. Upon reentering the app, everything updated as it should. Would appreciate additional testing in this area.
* I've reimplemented the dining logic following the old models closely. It's possible I missed something; we should pay careful attention and make sure I'm parsing hours right / determining if a venue is open or closed correctly.
* In general, everything seems to be working smoothly. But I'm of the opinion that this would be a great time to try out TestFlight to make sure.

## TODO
* Testflight
* @ josh somewhere along the line, I broke your HTML preloading for commons. Not sure the best place to add that in would be (func is still there, just not being called).


PennMobile/Dining/Model/DiningVenue+Extensions.swift
```diff
@@ -0,0 +1,108 @@
...
+        case .unknown: return "Other"
+        }
+    }
+}
+
+// MARK: - DiningVenue
+extension DiningVenue {
+    
+    // MARK: - Defaults
+    static let defaultVenueIds: [Int] = [593, 636, 1442]
```
joshdoman: What are these for? I thought we were aiming for a solution with zero data on the device.

PennMobile/Dining/Networking + Cache/DiningDataStore.swift
```diff
@@ -0,0 +1,56 @@
...
+
+class DiningDataStore {
+    
+    static let shared = DiningDataStore()
+    private var response: DiningAPIResponse = DiningAPIResponse(document: .init(venues: []))
+    private let dataStore: LocalJSONStore<DiningAPIResponse> = LocalJSONStore(storageType: .cache, filename: "venues.json")
+
+    private var todayString: String {
+        return Date.dayOfMonthFormatter.string(from: Date())
+    }
```
joshdoman: Let's keep pure date manipulation and in the Date Extension file.

PennMobile/Dining/Networking + Cache/LocalJSONStore.swift
```diff
@@ -0,0 +1,68 @@
...
+    }
+    
+    private var fileURL: URL {
+        return folder.appendingPathComponent(filename)
+    }
+    
+    private func ensureFolderExists() {
+        let fileManager = FileManager.default
+        var isDir: ObjCBool = false
+        if fileManager.fileExists(atPath: folder.path, isDirectory: &isDir) {
```
joshdoman: What is ObjCBool?

PennMobile/Dining/Networking + Cache/LocalJSONStore.swift
```diff
@@ -0,0 +1,68 @@
...
+        ensureFolderExists()
+    }
+    
+    // MARK: - Saving
+    func save(_ object: T) {
+        do {
+            let data = try JSONEncoder().encode(object)
+            try data.write(to: fileURL)
+        } catch let e {
+            print("CACHE ERROR: \(e)")
```
joshdoman: Let's remove this print statement. Do we want to handle failure?

PennMobile/Dining/Networking + Cache/LocalJSONStore.swift
```diff
@@ -0,0 +1,68 @@
...
+    var storedValue: T? {
+        guard FileManager.default.fileExists(atPath: fileURL.path) else {
+            return nil
+        }
+        do {
+            let data = try Data(contentsOf: fileURL)
+            return try JSONDecoder().decode(T.self, from: data)
+        } catch let e {
+            // This error will happen when our Codable representations change. This is okay, the cache will overwrite old structures eventually.
+            print("CACHE DECODING ERROR (CODABLE may have changed): \(e)")
```
joshdoman: Let's keep this print statement for debugging purposes.

PennMobile/Dining/Networking + Cache/StorageType.swift
```diff
@@ -0,0 +1,32 @@
...
+    
+    var searchPathDirectory: FileManager.SearchPathDirectory {
+        switch self {
+        case .cache: return .cachesDirectory
+        case .permanant: return .documentDirectory
+        }
+    }
+    
+    var folder: URL {
+        let path = NSSearchPathForDirectoriesInDomains(searchPathDirectory, .userDomainMask, true).first!
```
joshdoman: Any risk of this failing? I'm generally not a fan of bang operators.

PennMobile/Dining/Views/DiningCell.swift
```diff
@@ -75,8 +68,8 @@ extension DiningCell {
         }
     }
     
-    fileprivate func updateTimeLabel(with times: [OpenClose]?) {
-        timesLabel.text = times?.strFormat
+    fileprivate func updateTimeLabel(with venue: DiningVenue) {
+        timesLabel.text = venue.humanFormattedHoursStringForToday
```
joshdoman: Nice name.

PennMobile/Dining/Views/DiningCell.swift
```diff
@@ -47,19 +47,12 @@ class DiningCell: UITableViewCell {
...
-            titleLabel.text = DiningVenueName.getShortVenueName(for: venue.name)
-        } else {
-            titleLabel.text = venue.name.rawValue
-        }
-        
-        updateTimeLabel(with: venue.times)
-        
-        if let times = venue.times {
-            if times.isOpen {
+        if venue.hasMealsToday {
```
joshdoman: If the venue has no meals today, shouldn't it read Closed? But then how do we handle instances where we don't know and are awaiting a response from the API?

PennMobile/Home/Fitness/Controllers/FitnessViewController.swift
```diff
@@ -46,7 +46,7 @@ extension FitnessViewController {
         FitnessAPI.instance.fetchFitnessHours { (success, error) in
             DispatchQueue.main.async {
                 if !success {
-                    DiningHoursData.shared.clearHours()
+                    //DiningHoursData.shared.clearHours()
```
joshdoman: Can we remove this?

PennMobile/Home/Controllers/HomeViewController.swift
```diff
@@ -244,17 +244,17 @@ extension HomeViewController {
...
-    func saveSelection(for cafes: [DiningVenue]) {
+    func saveSelection(for venueIds: [Int]) {
         guard let diningItem = self.tableViewModel.getItems(for: [HomeItemTypes.instance.dining]).first as? HomeDiningCellItem else { return }
-        if cafes.count == 0 {
-            diningItem.venues = DiningVenue.getDefaultVenues()
+        if venueIds.count == 0 {
+            diningItem.venues = DiningDataStore.shared.getVenues(with: DiningVenue.defaultVenueIds)
         } else {
-            diningItem.venues = cafes
+            diningItem.venues = DiningDataStore.shared.getVenues(with: venueIds)
```
joshdoman: Looks good.

PennMobile/Home/Controllers/DiningCellSettingsController.swift
```diff
@@ -73,43 +56,26 @@ class DiningCellSettingsController: UITableViewController {
...
+        } else {
+            chosenVenueIds.insert(allVenues[indexPath.row].id)
         }
+        tableView.reloadRows(at: [indexPath], with: .automatic)
     }
 
     @objc func handleCancel() {
         self.dismiss(animated: true, completion: nil)
-        FirebaseAnalyticsManager.shared.trackEvent(action: .updateDiningPreferences, result: .cancelled, content: chosenCafes.count)
+        FirebaseAnalyticsManager.shared.trackEvent(action: .updateDiningPreferences, result: .cancelled, content: chosenVenueIds.count)
```
joshdoman: Let's remove the Firebase logs since they aren't relevant anymore. We track this information in the DB

PennMobile/Setup + Navigation/AppDelegate.swift
```diff
@@ -28,6 +28,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
     
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
+        
+        // Get the cache path for debugging purposes
+        print("Cache Path: \(DataStoreType.cache.folder)")
+
```
joshdoman: Let's delete this

PennMobile/Login/LoginWebviewController.swift
```diff
@@ -70,7 +70,7 @@ class LoginWebviewController: PennLoginController, IndicatorEnabled {
     
     private func getDiningBalance() {
         if let student = Student.getStudent(), student.isFreshman() {
-            DiningViewModel.ShowDiningPlan = true
+            DiningViewModel.showDiningPlan = true
```
joshdoman: I really really don't like this solution. It stems from us trying to ascertain whether a student is on a meal plan or not before we have actually tried to fetch their balance so that we can determine whether or not to show them the balance section on the dining page. Can we remove this line and change the dining page to show the balance section for all logged in users whether or not they have a balance?

DominicHolmes:
Great comments, thanks @joshdoman ! I'll work through those tonight.

> One concern, will this work on the dining card on the home page? The home page may try to load a dining card before we have any venue information. Do we handle that?
Yeah, so currently what happens is, the very first time the app is launched the dining venues array is empty. Nothing crashes, but the dining card is basically empty (just the title). I was thinking we could either:
1. Bundle a json file with the app that it decodes on first launch or 
2. Make a dining API request on first launch of app and fill out the array, before any UI is shown or
3. Something else

Thoughts?

PennMobile/Dining/Model/DiningVenue+Extensions.swift
DominicHolmes: These are ids for commons, hill, lch.

I figured we should still have a way to prioritize commons / provide defaults on the home page. We could also include this in our API endpoint if we really want to remove it from the code.

DominicHolmes: They're used to display the default selected dining halls

PennMobile/Dining/Networking + Cache/LocalJSONStore.swift
DominicHolmes: Docs for fileExists: https://developer.apple.com/documentation/foundation/filemanager/1410277-fileexists 

As far as I can tell, I can only check if the path is a directory with this function, which uses ObjCBool :/

PennMobile/Dining/Networking + Cache/LocalJSONStore.swift
DominicHolmes: Added a comment here explaining when it fails (when we screw up our encoding funcs).

I'm thinking we should leave this in, as otherwise we might silently break our Cache in the future.

DominicHolmes: I guess this could also fail when data.write fails, but we can't do anything about that. It's also just a cache, so okay if it fails to write once.. it'll just be overwritten next time.

Though if our encoding is messed up, it'll fail forever until we fix it.

PennMobile/Dining/Networking + Cache/LocalJSONStore.swift
DominicHolmes: agreed

PennMobile/Dining/Networking + Cache/StorageType.swift
DominicHolmes: got rid of it ^ unclear if this could fail

PennMobile/Dining/Views/DiningCell.swift
DominicHolmes: Exactly. We used to say "Closed Today" when there was no data, which was a bit confusing...

Possible fix would be to not filter out the "Closed" meal types?

PennMobile/Home/Controllers/DiningCellSettingsController.swift
DominicHolmes: Done

PennMobile/Login/LoginWebviewController.swift
DominicHolmes: So show balance for everyone logged in? Would that be Student.getStudent() to determine if they are logged in or not?

ccabo1:
Teh best kind of PR 
image

PennMobile/Login/LoginWebviewController.swift
DominicHolmes: Adding a GitHub issue, this will be in a different PR

DominicHolmes:
Made the following changes today:
- Fixed the "Closed" label bug
- Fixed the issue with Home page not reloading when the API response finishes on first load
- Fixed the same issue, but on DiningVC
- Filtered out Houston Market so that we don't have 2 sets of hours
- Added a "Opens in xH yM" label that displays when a venue is closed, but has an open meal coming that day

DominicHolmes:
@joshdoman @epoww this is ready for another review

PR #254, opened 2019-10-23
Remove ESTabBar pod

DominicHolmes:
Removing this pod because
1. Accessibility support is not great
2. It is involved with the selected tab bugs we've been experiencing

Everything seems to work correctly by just replacing the ESTabBarController instance with a UITabBarController.


joshdoman:
Screenshots? Does it highlight the icon of the tab your on?

DominicHolmes:
> Screenshots? Does it highlight the icon of the tab you're on?

Yep - I'm afk for a bit but it's visually totally identical. The big difference is no transition animations when tapping different tabs, but it looks like we aren't using those anyway.

PR #259, opened 2019-11-15
Transaction data

epoww:
Screen Shot 2019-11-15 at 6 00 42 PM


PennMobile/Dining/Controllers/DiningDollarsTransactionViewController.swift
```diff
@@ -7,3 +7,98 @@
...
+        tableView.delegate = self
+        tableView.dataSource = self
+                    
+        tableView.register(TransactionTableViewCell.self, forCellReuseIdentifier: TransactionTableViewCell.identifier)
+    }
+    
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        
+        PennCashNetworkManager.instance.getTransactionHistory { data in
```
DominicHolmes: Pull this into its own function

PennMobile/Dining/Controllers/DiningDollarsTransactionViewController.swift
```diff
@@ -7,3 +7,98 @@
...
+
+// MARK: - Table View Functions
+extension DiningDollarsTransactionViewController {
+    override func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        return transactionHistory?.count ?? 0
+    }
+    
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let cell = tableView.dequeueReusableCell(withIdentifier: TransactionTableViewCell.identifier, for: indexPath) as! TransactionTableViewCell
+        let transaction: Transaction = transactionHistory![indexPath.item]
```
DominicHolmes: Avoid force unwraps like this, make this handle the optional case with an empty cell

PennMobile/Dining/Controllers/DiningDollarsTransactionViewController.swift
```diff
@@ -7,3 +7,98 @@
...
+                completion(transactionAPIResponse.results, false)
+            } else {
+                completion(nil, true)
+            }
+        }
+    }
+}
+
+// MARK: - Transaction + Codable
+struct TransactionAPIResponse: Codable {
```
DominicHolmes: These should be in their own file

PennMobile/Dining/Views/DiningBalanceCell.swift
```diff
@@ -12,6 +12,7 @@ class DiningBalanceCell: UITableViewCell {
     
     static let identifier = "diningBalanceCell"
     static let cellHeight: CGFloat = 120
+    var viewController: DiningViewController?
```
DominicHolmes: Pretty sure this is going to create a strong reference cycle, @joshdoman can you confirm? We should use `weak` I think

PennMobile/Dining/Views/DiningViewModel.swift
```diff
@@ -73,6 +75,7 @@ extension DiningViewModel: UITableViewDataSource {
             let cell = tableView.dequeueReusableCell(withIdentifier: diningBalanceCell, for: indexPath) as! DiningBalanceCell
             cell.selectionStyle = .none
             cell.diningBalance = balance
+            cell.viewController = viewController
```
DominicHolmes: yeah definitely a strong cycle here

PennMobile/Dining/Views/TransactionTableViewCell.swift
```diff
@@ -7,3 +7,139 @@
 //
 
 import Foundation
+
+class TransactionTableViewCell: UITableViewCell {
+    
+    static let identifier = "transactionCell"
+    
+    var transaction: Transaction! {
```
DominicHolmes: does this need to be required? seems dangerous, would be better as optional

PennMobile/Dining/Views/TransactionTableViewCell.swift
```diff
@@ -7,3 +7,139 @@
...
+
+        descriptionLabel.leftAnchor.constraint(equalTo: leftAnchor, constant: 13).isActive = true
+        descriptionLabel.centerYAnchor.constraint(equalTo: centerYAnchor, constant: -10).isActive = true
+    }
+}
+
+// MARK: - Title Cased String Extension
+extension String {
+    func titlecased() -> String {
+        return self.replacingOccurrences(of: "([A-Z])", with: " $1",
```
DominicHolmes: dope

PennMobile/General/Networking + Analytics/UserDBManager.swift
```diff
@@ -169,10 +169,12 @@ extension UserDBManager {
 
 // MARK: - Transaction Data
 extension UserDBManager {
-    func saveTransactionData(csvStr: String) {
+    func saveTransactionData(csvStr: String, _ callback: (() -> Void)? = nil) {
         let url = "\(baseUrl)/dining/transactions"
         let params = ["transactions": csvStr]
         let request = getAnalyticsPostRequest(url: url, params: params)
-        sendRequest(request)
+        sendRequest(request) { (data, response, err) in
```
DominicHolmes: great way to handle this!

PennMobile/Assets.xcassets/Color/UI Palette/uiGroupedBackgroundSecondary.colorset/Contents 2.json
```diff
@@ -0,0 +1,28 @@
+{
```
DominicHolmes: Something funky going on here -- all of these files are duplicates... we will figure this out at the meeting. I was having similar problems last week

PennMobile/Dining/Views/DiningBalanceCell.swift
joshdoman: Yes, this will create a retain cycle and ```weak``` should be used. Moreover, you should never pass a full view controller to a cell. This is what delegate's are for.

PennMobile/Dining/Views/TransactionTableViewCell.swift
joshdoman: This is perfectly safe sine a cell will never be created without a transaction associated with it.

PennMobile/Dining/Controllers/DiningDollarsTransactionViewController.swift
joshdoman: This is perfectly safe and a rare instance where the bang operator is appropriate. If ```transactionHistory``` is null, then there will be zero cells by the numberOfCells delegate method and this delegate method will never be called.

PR #264, opened 2019-11-21
Submitting two-factor authentication code

henrYoda:
When using makeauthrequest (such as when fetching dining data), the program will now insert the user's two-factor code and submit it.


PennMobile/Auth/PennAuthRequestable.swift
```diff
@@ -152,4 +158,54 @@ extension PennAuthRequestable {
...
         task.resume()
     }
+
+    private func makeRequestWithTwoFac(targetUrl: String, shibbolethUrl: String, html: String, _ completionHandler: @escaping (Data?, URLResponse?, Error?) -> Void){
+        
+        guard let passcode = html.getMatches(for: "name=\"passcode\" value=\"(.*?)\"").first,
+        let required = html.getMatches(for: "name=\"required\" value=\"(.*?)\"").first,
+        let appfactor = html.getMatches(for: "name=\"appfactor\" value=\"(.*?)\"").first,
+        let ref = html.getMatches(for: "name=\"ref\" value=\"(.*?)\"").first,
+        let service = html.getMatches(for: "name=\"service\" value=\"(.*?)\"").first,
```
joshdoman: Are these fields necessary? I thought they were removed, which was why the previous auth flow stopped working.

henrYoda: Yeah, the two factor page is still submitting the post request using these fields for some reason.. its possible that they'll change this in the future.

PR #271, opened 2019-12-18
Remote Push Notifications

joshdoman:
Asks user to enable push notifications on log in and sends the device token to the server.
Screenshot 2019-12-17 at 9 16 41 PM


PennMobile/General/Networking + Analytics/UserDBManager.swift
```diff
@@ -32,6 +32,23 @@ class UserDBManager: NSObject, Requestable {
         let task = URLSession.shared.dataTask(with: request as URLRequest, completionHandler: callback)
         task.resume()
     }
+    
+    fileprivate func makePostRequestWithAccessToken(url: String, params: [String: String], callback: @escaping (Data?, URLResponse?, Error?) -> Void) {
```
DominicHolmes: Can you add a line comment above this function describing it? Like when it should be used vs. regular post requests, etc

joshdoman: Sure.

PennMobile/General/NotificationRequestable.swift
```diff
@@ -60,25 +72,24 @@ extension NotificationRequestable where Self: UIViewController {
...
                     // Checking for setting is opened or not
-                    print("Setting is opened: \(success)")
                     DispatchQueue.main.async {
                         completion?(success)
                     }
                 })
             }
         }
-        alertView.addButton("Decline", action: {  })
+        alertView.addButton("Decline", backgroundColor: UIColor(r: 114, g: 115, b: 117), action: {  })
```
DominicHolmes: Use a real color instead of hardcoding values pls

PennMobile/General/NotificationRequestable.swift
```diff
@@ -60,25 +72,24 @@ extension NotificationRequestable where Self: UIViewController {
...
     func alertForDetermination(_ completion: AuthorizedCompletion?) {
         let appearance = SCLAlertView.SCLAppearance(showCloseButton: false)
         let alertView = SCLAlertView(appearance: appearance)
-        alertView.addButton("Turn On"){
+        alertView.addButton("Turn On") {
             self.registerPushNotification(completion)
         }
-        alertView.addButton("Decline", action: {  })
-        alertView.showSuccess("Enable Notifications", subTitle: "Get notifications for laundry, and our future updates!")
+        alertView.addButton("Decline", backgroundColor: UIColor(r: 114, g: 115, b: 117), action: {  })
```
DominicHolmes: use color vars

PennMobile/Login/LoginController.swift
```diff
@@ -48,12 +48,20 @@ class LoginController: UIViewController, ShowsAlert {
...
 // MARK: - Login Completion Handler
-extension LoginController {
+extension LoginController: NotificationRequestable {
     func loginCompletion(_ successful: Bool) {
         if successful {
             // Login Successful
             UserDefaults.standard.setLastLogin()
             AppDelegate.shared.rootViewController.switchToMainScreen()
+            
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```
DominicHolmes: We should do this on a non-main thread. If the Notification Center stuff wants to happen on a main thread, we can put that inside `self.requestNotification()`

joshdoman: What exactly are you suggesting? This is just a 0.5 second delay. Notification Center stuff still happens on the background thread.

joshdoman: All this is saying is asynchronously wait 0.5 seconds and then from the Main thread, call requestNotification(). Request notification then asks Notification Center if a notification has been enabled, which then responds on a background thread "yes" or "no."

DominicHolmes: just suggesting the work gets offloaded to a background thread, since it's not UI based or user focused.

so that would be `DispatchQueue.global(qos: .background).asyncAfter(deadline: .now() + 0.5)`

DominicHolmes: main queue handles UI and keeps things 60fps, so anytime we can async on a background queue we should

joshdoman: Implemented.

PennMobile/General/NotificationRequestable.swift
joshdoman: Implemented.

PennMobile/General/NotificationRequestable.swift
joshdoman: Implemented.

PR #272, opened 2019-12-20
Fixing Dining Balance Display Bug

joshdoman:
This PR fixes the dining dollar bug that was causing the app to show some user's balance. All users now see the dining balance UI, even if the app determines they do not have a balance. This fixes the previous issue, which was caused by the app incorrectly inferring whether or not someone had a balance. Note that if the app knows a user doesn't have a balance, it does not proactively attempt to fetch a new one from CampusExpress unless explicitly told to refresh.

image

### DIFF
--- a/PennMobile/Dining/Controllers/DiningViewController.swift
+++ b/PennMobile/Dining/Controllers/DiningViewController.swift
@@ -35,13 +35,17 @@ class DiningViewController: GenericTableViewController {
         super.viewWillAppear(animated)
         fetchDiningHours()
         
-        if viewModel.shouldShowDiningBalances {
+        UserDefaults.standard.set(hasDiningPlan: false)
+        if UserDefaults.standard.hasDiningPlan() {
             if viewModel.balance == nil {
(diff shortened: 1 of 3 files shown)


PennMobile/Dining/Controllers/DiningViewController.swift
```diff
@@ -35,13 +35,17 @@ class DiningViewController: GenericTableViewController {
         super.viewWillAppear(animated)
         fetchDiningHours()
         
-        if viewModel.shouldShowDiningBalances {
+        UserDefaults.standard.set(hasDiningPlan: false)
```
DominicHolmes: are we explicitly setting this to false and then immediately doing an if check?

DominicHolmes:
rest of the pr looks good to me!

PennMobile/Dining/Controllers/DiningViewController.swift
joshdoman: Oops meant to remove that.

PR #274, opened 2019-12-21
Automatic Debug Mode

joshdoman:
This PR simplifies the release process by removing the need to manually change the codebase for production. Previously, certain flags had to be changed to `false`, such as `FeedAnalyticsManager.shared.dryrun` and certain pieces of code had to be commented out, such as `FirebaseConfiguration.shared.setLoggerLevel(.min)`. This is now handled automatically with `#if Debug`, which is a new feature available in iOS 10.

Example:
```
#if DEBUG
    FirebaseConfiguration.shared.setLoggerLevel(.min)
#endif
``` 
**NOTE:** New releases still require a change in the build schema and an update to the build and version numbers.


DominicHolmes:
super super approved

PR #286, opened 2020-01-01
Courses privacy implementation

DominicHolmes:
* Integrated SwiftUI screen to request permission to upload courses.
* Added UserDefaults fields that record when we last asked permission for each privacy option.
* Added UserDefaults fields that record we last shared data corresponding to each privacy option.
* Added a UserDBManager function that uploaded anonymized courses to account/courses using the new OAuth functions.

Still need to:
* Add a storage mechanism for the user's on-device UUID which allows them to upload courses as an anonymous user while maintaining our ability to delete their data at their request. This should use a combination of Keychain-based UUID (with iCloud storage), and a hashing of the user's pennkey and password. The idea is that both of these wouldn't change at once -- both are relatively stable storage solutions but have a good chance of changing over the course of 4 years.
* Actually call the privacy request screen (not sure where to do this from)


PennMobile/About/Privacy/PermissionView.swift
```diff
@@ -0,0 +1,133 @@
...
+        delegate.userDecision = .affirmative
+    }
+    
+    func doNotShareCourses() {
+        delegate.userDecision = .negative
+    }
+    
+    func closeView() {
+        delegate.userDecision = .close
+    }
```
joshdoman: When is this triggered?

DominicHolmes: its not -- was going to add an X button then decided against it

PR #287, opened 2020-01-03
College House Privacy-Permission

joshdoman:
This PR implements college house privacy permission. If permission is denied, all college house data is deleted from the server. Likewise, if permission is granted, the user's college house is retrieved from CampusExpress and saved.


joshdoman:
We're not asking for it anywhere right now except on the Privacy Screen.

PR #288, opened 2020-01-03
Academic Profile Privacy-Permission

joshdoman:
This PR implements academic profile privacy permission. If permission is denied, all academic profile data (school, year, etc.) is deleted from the server. Likewise, if permission is granted, the user's academic profile is retrieved and saved.

In addition, this PR contains login and privacy screen improvements. Namely, each PrivacyOption now  has a `requiresAuth` field, which if true, requires the user to sign-in with their PennKey before giving Penn Mobile permission to their data. Sign-in is never required when removing permission. Not only does this give greater security, it also ensures that 2FA is not required when accessing their information.


DominicHolmes:
Awesome work! Looks great.

PR #294, opened 2020-01-18
Merge "manage-groups-2" (Group-GSR-Development) into "development"

r3khaan:
Hid the "groups" tab on the GSR Booking screen, so the group gsr development is all hidden.


PennMobile/GSR-Booking/Controllers/GSRGroupNewIntialController.swift
```diff
@@ -0,0 +1,284 @@
...
+        colorCollectionView.topAnchor.constraint(equalTo: colorLabel.bottomAnchor, constant: 20).isActive = true
+        colorCollectionView.leftAnchor.constraint(equalTo: view.leftAnchor, constant: 30).isActive = true
+        colorCollectionView.rightAnchor.constraint(equalTo: view.rightAnchor, constant: -20).isActive = true
+        colorCollectionView.heightAnchor.constraint(equalToConstant: 40).isActive = true
+        colorCollectionView.translatesAutoresizingMaskIntoConstraints = false
+    }
+
+
+    @objc func createGroupBtnAction(sender:UIButton!) {
+        GSRGroupNetworkManager.instance.createGroup(name: nameField.text!, color: colorLabel.text!) { (success, errorMsg) in
```
joshdoman: Let's try to avoid bang operators unless it's obviously guaranteed to never be null. Will this crash if the user presses create group without choosing the name or color?

DominicHolmes:
Don’t merge this until I get a chance to look at it @josh there were lots of xcodeproj file changes we should be careful of

PennMobile/GSR-Booking/Controllers/GSRGroupNewIntialController.swift
r3khaan: Not my code, but currently the button gets disabled until the user selects a name & color so that they will not be null. I can still fix this in a commit though.

PennMobile/GSR-Booking/Model/GSRGroup.swift
```diff
@@ -8,18 +8,63 @@
...
+    let userSettings: GSRGroupIndividualSettings? //not optional, beacuse we need to know if pennKey is Active
+
+    var imgURL: String?
+    var owners: [GSRGroupMember]?
+    var members: [GSRGroupMember]?
+    let reservations: [String]? //array of reservationID's
+    let groupSettings: GSRGroupAccessSettings?
+    
+    static let groupColors: [String : UIColor] = [
+        "Labs Blue" : UIColor(red: 32, green: 156, blue: 238),
```
DominicHolmes: make array of our custom color type, not explicitly defined

DominicHolmes:
I only reviewed the pbxproj stuff but that part looks good. Deferring to josh on code approval

PR #298, opened 2020-02-09
Dining Stats

epoww:
Put old dining view and new blank statistics view in a tab bar controller (like the booking & reservations views on GSR page)


joshdoman:
@epoww Can you attach a screenshot? We should also add a switch or flag so that the current home page (no tab bar) is rendered in production until the statistics view is finished.

epoww:
Here's a screen shot of the tab bar

Simulator Screen Shot - iPhone 11 - 2020-02-14 at 17 48 00

DominicHolmes:
@01jongmin

DominicHolmes:
TODO: fix googleserviceinfo.plist changes (something broke w references)

DominicHolmes:
there are 29 pbxproj conflicts o.O

might try to merge those with an automatic script

PennMobile/Dining/Controllers/Statistics/DiningInsightsViewController.swift
```diff
@@ -30,17 +30,21 @@ class DiningInsightsViewController: GenericViewController {
...
+            case .failure(.authenticationError):
+                //TODO: handle authenticationError
+                print("reauthentication?")
+            case .failure(.noInternet):
+                DispatchQueue.main.async { self.navigationVC?.addStatusBar(text: .noInternet) }
+            default:
+                DispatchQueue.main.async { self.navigationVC?.addStatusBar(text: .apiError) }
             }
+            
+            self.diningInsights = DiningAPI.instance.getCachedDiningInsights()
```
DominicHolmes: can we do this only when there is a failure? it's a little confusing that we do this all the time (even though it works). we should use the result if it was .success.

01jongmin: I thought that the cache exist such that even if a user, for example, opens their dining insights tab without the internet, if they had access to the internet and cached the result of their dining insights previously, they would still be able to view it.

DominicHolmes: resolved

Issue #280, opened 2019-12-27
Add rich notification support for GSR bookings


DominicHolmes:
- Add a photo of the GSR, if available
- Add a button to cancel the reservation

PR #308, opened 2020-02-15
3 claws

DominicHolmes:
#280 

Adding support for rich notification payloads. Not implemented yet: the cancel button or share button actions. They will show up but are not functional

 This PR will also require modification of notification payload structure, like so:

`EXAMPLE NOTIFICATION PAYLOAD:
     {
         "aps": {
             "alert": {
                 "body": "This is the body text, concerning your GSR in WIC RM 123 from 9:30-11am.",
                 "title": "Upcoming GSR",
                 "subtitle": "WIC Rm 129",
                 "other info, like booking ids": "etc"
             },
             "mutable-content": 1,
             "thread-id": 1,
             "category": "UPCOMING_GSR"
         },
         "media-url": "https://GSR-PICTURE-URL-HERE",
         "sound": "default"
     }`


joshdoman:
@DominicHolmes Can you attach a screenshot?

DominicHolmes:
@joshdoman would have to set up the notification payload stuff to simulate it

PennMobile/Notifications/Rich_Notifications/GSRNotificationExtension.swift
```diff
@@ -0,0 +1,108 @@
...
+                 "title": "Upcoming GSR",
+                 "subtitle": "WIC Rm 129"
+             },
+             "mutable-content": 1,
+             "thread-id": 1,
+             "category": "UPCOMING_GSR"
+         },
+         "media-url": "https://snworksceo.imgix.net/dpn/20c5f2b4-9f36-4c47-985b-5b9673f86d77.sized-1000x1000.jpg?w=1000",
+         "sound": "default"
+     }
```
joshdoman: From [this documentation](https://developer.apple.com/library/archive/documentation/NetworkingInternet/Conceptual/RemoteNotificationsPG/CreatingtheNotificationPayload.html), it seems like custom data should be passed through a key containing `acme` in it. Is that correct? If so, I'm thinking I pass you the reservation info, which includes the booking ID, there.

DominicHolmes: I think custom keys can just be specified at the top level. ‘acme’ can be any key.

I’m already using “media-url” as a custom key.

DominicHolmes: Would be great to have the res id and necessary stuff to cancel the booking, and also the room name etc

DominicHolmes:
Completed feature, added functionality to Cancel and Share bookings from the rich notification:

Screen Shot 2020-03-06 at 8 16 53 PM

Screen Shot 2020-03-06 at 8 17 07 PM

DominicHolmes:
@joshdoman pls review

PR #309, opened 2020-02-18
Light Lunch

joshdoman:
This PR modifies the status label to say "Open for Light Lunch" instead of "Open" during light hours. Light lunch is a priority for dining services, and they are not happy at all that we don't have it on the app and show the dining halls as closed. Adding light lunch is something we must do if we want to get back in their good graces and collaborate with them in the future. In another PR, we should add a clarifier or a second row for light lunch times to differentiate them from regular hours. 

image


PennMobile/Dining/Views/DiningCell.swift
```diff
@@ -53,10 +53,17 @@ extension DiningCell {
         updateTimeLabel(with: venue)
         
         if venue.hasMealsToday {
-            if venue.isOpen {
-                statusLabel.text = "Open"
-                statusLabel.textColor = .baseYellow
-                statusLabel.font = .primaryInformationFont
+            if let meal = venue.currentMeal {
+                if meal.isLight {
+                    // Label will read: "Serving Light Lunch" or "Serving Light Breakfast"
```
DominicHolmes: I think this comment is outdated

DominicHolmes:
Also I think a slightly more concise version would be just "Light Lunch". As its highlighted yellow I think ppl will assume its open.

Other than that, LGTM!

PR #310, opened 2020-02-18
Home Page Dining Hours Bug Fix

joshdoman:
This PR should fix the bug that prevented dining hours from being updated on the home page. This bug was only evident on Sundays when the dining halls were displayed  as "Closed" on the home page even though they were open in the dining tab. The issue was that the venues in the home page item contained the hours for the previous week and were only being updated if the venues had not yet been fetched from the internal model.


DominicHolmes:
Could you comment more on why we're using this new _getUpdatedHours_ function only on the home page? It seems like this would be a problem on DiningVC too.

If you could leave additional comments in the code above this function that would be great. Specifically when you should use this rather than `getVenues()`

Other than that looks great! Can't believe u caught this bug. That's tricky!

joshdoman:
> Could you comment more on why we're using this new _getUpdatedHours_ function only on the home page? It seems like this would be a problem on DiningVC too.
> 
> If you could leave additional comments in the code above this function that would be great. Specifically when you should use this rather than `getVenues()`
> 
> Other than that looks great! Can't believe u caught this bug. That's tricky!

The only difference is that it maintains the order of the venues that you're requesting.

joshdoman:
> Could you comment more on why we're using this new _getUpdatedHours_ function only on the home page? It seems like this would be a problem on DiningVC too.
> 
> If you could leave additional comments in the code above this function that would be great. Specifically when you should use this rather than `getVenues()`
> 
> Other than that looks great! Can't believe u caught this bug. That's tricky!

I've removed the other `getVenues(for: ids)` functions so now there are only two. `getVenues()` returns all the venues and `getVenues(for venueIds: [Int])`

PR #311, opened 2020-02-19
Asking the user to fetch Two-Step code again if failed last time

henrYoda:
image

This will make Penn Mobile ask whether the user wants to try and fetch his Two-Step code again if he enabled it but the app failed to fetch it last time (perhaps he closed the app too early?). The program will only ask this once before disabling Two-Step if it fails to fetch the code again.

I also added some extensions to UserDefaults pertaining to values used in Two-Step code fetching.

This also contains some bug fixes pertaining to fetching the code on login as opposed to from the more tab.

@joshdoman Can you check whether this PR breaks notification permission requesting? I was not able to test this as I can't run it on my device


test.cpp
```diff
@@ -0,0 +1,16 @@
...
+    int count = 0;
+
+    for(int i = 1; i <= n/2; i*=2){
+	for(int j = i; j < 4*i; j*=2) {
+	    count++;
+	}
+    }
+
+    cout << count << endl;
+}
```
joshdoman: Umm what's this?

PennMobile/Setup + Navigation/RootViewController.swift
```diff
@@ -373,4 +391,24 @@ extension RootViewController : TwoFactorEnableDelegate {
             askForNotificationPermission()
         }
     }
+    ///Checks whether the user has enabled Two Factor but the code was never fetched. This can happen when the user closes the
+    ///app before it is done fetching the code. Don't try to fetch it if Two Factor was enabled more than a week ago. Also, don't try to fetch
+    ///it again if the user already tried again once.
```
joshdoman: Use multi-line documentation / a block comment. See [here](https://sarunw.com/posts/swift-documentation/#syntax) for example.

PennMobile/Setup + Navigation/RootViewController.swift
```diff
@@ -133,6 +128,25 @@ class RootViewController: UIViewController, NotificationRequestable {
...
+                let alert = UIAlertController(title: "Two-Step Code Not Fetched", message: "We were unable to fetch your Two-Step code last time. Do you want to try again?", preferredStyle: .alert)
+                alert.addAction(UIAlertAction(title: "No", style: .default, handler: { _ in
+                    alert.dismiss(animated: true, completion: nil)
+                    UserDefaults.standard.setTwoFactorEnabled(to: false);
+                }))
+                
+                alert.addAction(UIAlertAction(title: "Yes", style: .default, handler: { _ in
+                    let twc = TwoFactorWebviewController()
+                    UserDefaults.standard.setTwoFactorEnabledAgain(to: true)
+                    self.current.present(twc, animated: true)
```
joshdoman: Why are we showing a webview controller here instead of just making a request directly? Do we need them to login again?

PennMobile/Setup + Navigation/RootViewController.swift
```diff
@@ -133,6 +128,25 @@ class RootViewController: UIViewController, NotificationRequestable {
             
         }
         
+        //Request to fetch Two Factor Code again if the app failed to fetch it last time
+        if shouldFetchTwoFactorCode() {
```
joshdoman: We need to re-architect how we display pop-ups. Right now, we have a series of if statements, each individually checking if it should display its popup. In the short-term, we should maybe make them if-else statements so that no more than one gets run on a single app open. My concern is that the two-factor code popup will show up on top of the courses privacy screen.

PennMobile/Setup + Navigation/RootViewController.swift
```diff
@@ -133,6 +128,25 @@ class RootViewController: UIViewController, NotificationRequestable {
...
+            DispatchQueue.main.async {
+                let alert = UIAlertController(title: "Two-Step Code Not Fetched", message: "We were unable to fetch your Two-Step code last time. Do you want to try again?", preferredStyle: .alert)
+                alert.addAction(UIAlertAction(title: "No", style: .default, handler: { _ in
+                    alert.dismiss(animated: true, completion: nil)
+                    UserDefaults.standard.setTwoFactorEnabled(to: false);
+                }))
+                
+                alert.addAction(UIAlertAction(title: "Yes", style: .default, handler: { _ in
+                    let twc = TwoFactorWebviewController()
+                    UserDefaults.standard.setTwoFactorEnabledAgain(to: true)
```
joshdoman: I'm wary of us trying to manage so many states because it will make it hard to debug in the future. Rather than having another user defaults variable for "enabled again," why don't we just set "twoFactorEnabled" to false?

PennMobile/General/UserDefaults + Helpers.swift
```diff
@@ -508,3 +508,34 @@ extension UserDefaults {
...
+        UserDefaults.standard.set(newValue, forKey: "TOTPEnabled")
+        if newValue {
+            setTwoFactorEnabledDate(Date())
+        } else {
+            setTwoFactorEnabledAgain(to: false)
+        }
+    }
+    
+    func getTwoFactorEnabled() -> Bool {
+        return UserDefaults.standard.bool(forKey: "TOTPEnabled")
```
joshdoman: Let's place this key inside UserDefaultsKeys. The main reason is that this will cause the key to be removed on log out. We wouldn't want to ask a logged out user to re-enabled 2FA anyways. In addition, from a code quality perspective, it's almost always a good idea to put repeated strings in a variable name to avoid making mistakes in the future.

test.cpp
henrYoda: Oops this was something I was using to test something totally unrelated I did not realize I saved it in the penn mobile repository LOL

PennMobile/Setup + Navigation/RootViewController.swift
henrYoda: This is something you had mentioned - that we don't want to fetch the user's code without making them login when they reopen the app - to make it clear that we have the user's permission.

PennMobile/General/UserDefaults + Helpers.swift
henrYoda: Done!

PR #317, opened 2020-03-07
Pac code

01jongmin:
Enable users to view their PAC Code if they are logged in. 
Users can enable FaceID to authenticate themselves the first time they access PAC code data
Requires authentication using biometrics each time they access the data.
Has a refresh button on the top right screen to refetch PAC Code data.
Stores PAC Code data in the user's keychain, and removes it when the user intentionally/unintentionally logs out. 

Throws an alert if user attempt to access PAC Code without logging in.

For security reasons, I subtracted a random 4 digit number from my PAC Code when taking the screen shots below.

Created a "LocalAuthentication" protocol that enables users to use biometrics/password to access specific information

Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-07 at 10 26 20

Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-07 at 10 26 26

Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-07 at 10 26 31

Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-07 at 10 26 41

### DIFF
--- a/PennMobile/Auth/PennAuthRequestable.swift
+++ b/PennMobile/Auth/PennAuthRequestable.swift
@@ -24,6 +24,12 @@ extension PennAuthRequestable {
         let url = URL(string: targetUrl)!
         let request = URLRequest(url: url)
         let task = URLSession.shared.dataTask(with: request) { (data, response, error) in
+            
+            if let error = error, (error as NSError).code == -1009 {
+                completionHandler(nil, nil, NetworkingError.noInternet)
+                return
+            }
+            
             if let urlStr = response?.url?.absoluteString, urlStr == targetUrl {
                 UserDefaults.standard.setShibbolethAuth(authedIn: true)
                 completionHandler(data, response, error)
@@ -43,8 +49,7 @@ extension PennAuthRequestable {
                 }
             } else {
                 UserDefaults.standard.setShibbolethAuth(authedIn: false)
-//                completionHandler(nil, nil, NetworkingError.authenticationError)
-                completionHandler(nil, nil, NetworkingError.noInternet)
+                completionHandler(nil, nil, NetworkingError.other)
             }
             UserDefaults.standard.storeCookies()
         }
--- a/PennMobile/Auth/PennAuthRequestable.swift
+++ b/PennMobile/Auth/PennAuthRequestable.swift
@@ -43,7 +43,8 @@ extension PennAuthRequestable {
                 }
             } else {
                 UserDefaults.standard.setShibbolethAuth(authedIn: false)
-                completionHandler(nil, nil, NetworkingError.authenticationError)
+//                completionHandler(nil, nil, NetworkingError.authenticationError)
+                completionHandler(nil, nil, NetworkingError.noInternet)
             }
             UserDefaults.standard.storeCookies()
         }
--- a/PennMobile/Auth/PennLoginController.swift
+++ b/PennMobile/Auth/PennLoginController.swift
@@ -147,9 +147,11 @@ class PennLoginController: UIViewController, WKUIDelegate, WKNavigationDelegate,
         }
     }
     
+    var handleCancel: (() -> Void)? = nil
+    
     @objc fileprivate func cancel(_ sender: Any) {
         _ = self.resignFirstResponder()
-        dismiss(animated: true, completion: nil)
+        dismiss(animated: true, completion: self.handleCancel)
     }
     
     // MARK: - Decide Policy Upon Completed Navigation
--- a/PennMobile/General/Protocols.swift
+++ b/PennMobile/General/Protocols.swift
@@ -147,14 +147,14 @@ extension LocationPermissionRequestable {
     }
 }
 
-protocol LocalAuthentication {
+protocol LocallyAuthenticatable {
     func handleAuthenticationSuccess()
     func handleAuthenticationFailure()
 }
 
-extension LocalAuthentication {
+extension LocallyAuthenticatable {
     
-    func handleAuthentication(cancelText : String, reasonText: String) {
+    func requestAuthentication(cancelText : String, reasonText: String) {
         let context = LAContext()
         context.localizedCancelTitle = cancelText
 
--- a/PennMobile/General/Protocols.swift
+++ b/PennMobile/General/Protocols.swift
@@ -8,6 +8,7 @@
 
 import MBProgressHUD
 import CoreLocation
+import LocalAuthentication
 
 protocol IndicatorEnabled {}
 
@@ -142,6 +143,44 @@ extension LocationPermissionRequestable {
             }
         } else {
             return false
+        }
+    }
+}
+
+protocol LocalAuthentication {
+    func handleAuthenticationSuccess()
+    func handleAuthenticationFailure()
+}
+
+extension LocalAuthentication {
+    
+    func handleAuthentication(cancelText : String, reasonText: String) {
+        let context = LAContext()
+        context.localizedCancelTitle = cancelText
+
+        // Check if we have the needed hardware support.
+        var error: NSError?
+        if context.canEvaluatePolicy(.deviceOwnerAuthentication, error: &error) {
+            context.evaluatePolicy(.deviceOwnerAuthentication, localizedReason: reasonText ) { success, error in
+
+                if success {
+                    // Move to the main thread because the user may request UI changes.
+                    DispatchQueue.main.async {
+                        self.handleAuthenticationSuccess()
+                    }
+
+                } else {
+                    // Failed to authenticate with FaceID/passcode
+                    DispatchQueue.main.async {
+                        self.handleAuthenticationFailure()
+                    }
+                }
+            }
+        } else {
+            // Can't evaluate policy
+            DispatchQueue.main.async {
+                self.handleAuthenticationFailure()
+            }
         }
     }
 }
--- a/PennMobile/New Group/PAC Code/PacCodeNetworkManager.swift
+++ b/PennMobile/New Group/PAC Code/PacCodeNetworkManager.swift
@@ -23,21 +23,24 @@ extension PacCodeNetworkManager: PennAuthRequestable {
         return "https://penncard.apps.upenn.edu/penncard/jsp/fast2.do/Shibboleth.sso/SAML2/POST"
     }
     
-    func getPacCode(callback: @escaping ((_ code: String?) -> Void)) {
+    func getPacCode(callback: @escaping (_ result: Result<String, NetworkingError>) -> Void ) {
         makeAuthRequest(targetUrl: pacURL, shibbolethUrl: shibbolethUrl) { (data, response, error) in
-            if let httpResponse = response as? HTTPURLResponse {
-                if httpResponse.statusCode == 200 {
-                    if let data = data, let html = NSString(data: data, encoding: String.Encoding.utf8.rawValue) as String? {
-                        do {
-                            let pacCode = try self.findPacCode(from: html)
-                            return callback(pacCode)
-                        } catch {
-                            return callback(nil)
-                        }
-                    }
+            
+            guard let data = data, let html = NSString(data: data, encoding: String.Encoding.utf8.rawValue) else {
+                if let error = error as? NetworkingError {
+                    callback(.failure(error))
+                } else {
+                    callback(.failure(.other))
                 }
-            } else {
-                return callback(nil)
+                return
+            }
+            
+            do {
+                let pacCode = try self.findPacCode(from: html as String)
+                return callback(.success(pacCode))
+            } catch {
+                print("parsing error")
+                return callback(.failure(.parsingError))
             }
         }
     }
@@ -62,8 +65,16 @@ extension PacCodeNetworkManager: PennAuthRequestable {
             throw NetworkingError.parsingError
         }
         
-        // PAC Code is stored in the 5th index of the array
-        return identity[5]
+        
+        if (identity.count == 6) {
+            
+            // PAC Code is stored in the 5th index of the array
+            return identity[5]
+        } else {
+            throw NetworkingError.parsingError
+        }
+        
+        
     }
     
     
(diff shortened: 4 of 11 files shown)


PennMobile/New Group/PAC Code/PacCodeNetworkManager.swift
```diff
@@ -0,0 +1,71 @@
...
+                    let colContent = try col.text()
+                    identity.append(colContent)
+                }
+            }
+        } catch {
+            throw NetworkingError.parsingError
+        }
+        
+        // PAC Code is stored in the 5th index of the array
+        return identity[5]
```
joshdoman: This isn't safe and would crash the app if they changed the website so that there are less than 5 indices in the array. To be safe, let's check the length of the array before indexing.

PennMobile/New Group/PAC Code/PacCodeViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+        pacCodeInfoLabel.font = UIFont.systemFont(ofSize: 35)
+        pacCodeInfoLabel.text = "Your PAC Code is"
+        
+        view.addSubview(pacCodeInfoLabel)
+        pacCodeInfoLabel.translatesAutoresizingMaskIntoConstraints = false
+        pacCodeInfoLabel.bottomAnchor.constraint(equalTo: pacCodeHStack.topAnchor, constant: -30).isActive = true
+        pacCodeInfoLabel.centerXAnchor.constraint(equalTo: view.safeAreaLayoutGuide.centerXAnchor).isActive = true
+    }
+    
+    func refreshPacCode() {
```
joshdoman: This function is a bit of a misnomer. When I read `refreshPacCode`, I imagine that we're actually refreshing it from the CampusExpress. Let's change it to something else, like `setPACCode` or `updatePACCode`.

PennMobile/General/Protocols.swift
```diff
@@ -145,3 +146,41 @@ extension LocationPermissionRequestable {
...
+
+extension LocalAuthentication {
+    
+    func handleAuthentication(cancelText : String, reasonText: String) {
+        let context = LAContext()
+        context.localizedCancelTitle = cancelText
+
+        // Check if we have the needed hardware support.
+        var error: NSError?
+        if context.canEvaluatePolicy(.deviceOwnerAuthentication, error: &error) {
```
joshdoman: Under what circumstance would the user not have the needed hardware support?

PennMobile/General/Protocols.swift
```diff
@@ -145,3 +146,41 @@ extension LocationPermissionRequestable {
         }
     }
 }
+
+protocol LocalAuthentication {
+    func handleAuthenticationSuccess()
+    func handleAuthenticationFailure()
+}
+
+extension LocalAuthentication {
```
joshdoman: Nice protocol extension!

PennMobile/General/Protocols.swift
```diff
@@ -145,3 +146,41 @@ extension LocationPermissionRequestable {
         }
     }
 }
+
+protocol LocalAuthentication {
```
joshdoman: Let's change this protocol name so that it's not the same as the Apple Package. Maybe `LocallyAuthenticatable` or something else that reads nicely.

PennMobile/New Group/PAC Code/PacCodeViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+        super.viewDidDisappear(animated)
+        pacCode = nil
+        refreshPacCode()
+    }
+    
+    func handleAuthenticationSuccess() {
+        if let pacCode = getPacCode() {
+            self.pacCode = pacCode
+        } else {
+            self.showAlert(withMsg: "Please login to use this feature", title: "Login Error", completion: { self.navigationController?.popViewController(animated: true)} )
```
joshdoman: This alert doesn't make much sense. The user has successfully authenticated with FaceID or passcode but the pac code isn't there. That doesn't mean the user isn't logged in though. It just means we don't have the passcode stored locally on the device. We should be fetching the passcode from CampusExpress here instead.

PennMobile/New Group/PAC Code/PacCodeViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+    
+    fileprivate func handleNetworkPacCodeRefreshCompletion(_ pacCode: String?) {
+        DispatchQueue.main.async {
+            self.hideActivity()
+            if let pacCode = pacCode {
+                self.savePacCode(pacCode)
+                self.pacCode = pacCode
+                self.showAlert(withMsg: "Your PAC Code has been updated.", title: "Success!", completion: self.refreshPacCode)
+            } else {
+                self.showAlert(withMsg: "Unable to access your courses. Please try again later.", title: "Uh oh!", completion: nil)
```
joshdoman: Huh? I think you meant to change this.

PennMobile/New Group/PAC Code/PacCodeNetworkManager.swift
```diff
@@ -0,0 +1,71 @@
...
+    }
+    
+    func getPacCode(callback: @escaping ((_ code: String?) -> Void)) {
+        makeAuthRequest(targetUrl: pacURL, shibbolethUrl: shibbolethUrl) { (data, response, error) in
+            if let httpResponse = response as? HTTPURLResponse {
+                if httpResponse.statusCode == 200 {
+                    if let data = data, let html = NSString(data: data, encoding: String.Encoding.utf8.rawValue) as String? {
+                        do {
+                            let pacCode = try self.findPacCode(from: html)
+                            return callback(pacCode)
```
joshdoman: We want to start using Result types in the app instead of optional values. Here is a [link](https://www.hackingwithswift.com/articles/161/how-to-use-result-in-swift) to a guide for how that works.  @henrYoda implemented this in his current PR, which you can see [here](https://github.com/pennlabs/penn-mobile-ios/blob/automaticlogin/PennMobile/Auth/PlatformNetworkManager.swift#L28). The idea is we want to pass back the `NetworkingError`, which will contain the reason for failure (parsing error, authentication error, etc.). If it's an authentication error, we want to popup a login controller so they can log back in, and if it's a parsing error, we should throw an alert saying that it was unable to fetch your code.

PennMobile/New Group/PAC Code/PacCodeViewController.swift
joshdoman: Refer to my previous comments about passing back the error type. If we failed to get the pac code because the user's session has expired, we should ask them to log back into the app before giving up.

PennMobile/New Group/PAC Code/PacCodeViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+    }
+}
+
+// MARK: - Local Authentication
+extension PacCodeViewController : LocalAuthentication {
+        
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        
+        handleAuthentication(cancelText: "Go Back", reasonText: "Authenticate to see your PAC Code")
```
joshdoman: This function is a bit of misnomer. You're not handling authentication, you're requesting authentication. Let's change this to reflect that.

PennMobile/General/Protocols.swift
01jongmin: It is possible for users to disable both Touch ID and FaceID on their device. The `.canEvaluatePolicy` checks if the we can access biometrics before calling the  `canEvaluatePolicy(_:error:)` that might crash the app without the check. It is mentioned in the **Discussion** portion of this [documentation](https://developer.apple.com/documentation/localauthentication/lacontext/1514149-canevaluatepolicy).

PennMobile/New Group/PAC Code/PacCodeViewController.swift
01jongmin: Sorry! I should have caught this mistake.

PennMobile/New Group/PAC Code/PacCodeViewController.swift
```diff
@@ -135,16 +143,52 @@ extension PacCodeViewController {
...
+                self.showAlert(withMsg: "Something went wrong.\nPlease try again later.", title: "Uh oh!", completion: { self.navigationController?.popViewController(animated: true) } )
+            }
+        }
+    }
+    
+    fileprivate func handleAuthenticationError() {
+        
+        let llc = LabsLoginController { (success) in
+            DispatchQueue.main.async {
+                print("success")
```
joshdoman: Don't forget to remove this before merging.

PennMobile/New Group/PAC Code/PacCodeViewController.swift
```diff
@@ -86,28 +86,36 @@ class PacCodeViewController : UIViewController, KeychainAccessible, ShowsAlert,
...
+                // Handle the case in which the user is logged in but hasn't yet fetched their PAC Codes
+                // Acts as though the user pressed the refresh button
+                PacCodeNetworkManager.instance.getPacCode { result in self.handleNetworkPacCodeRefreshCompletion(result) }
+            } else {
+                self.showAlert(withMsg: "Please login to use this feature", title: "Login Error", completion: { self.navigationController?.popViewController(animated: true)} )
+            }
         }
         
-        refreshPacCode()
+        updatePACCode()
```
joshdoman: Let's only call this function if the pacCode is not nil. No reason to call it otherwise.

PennMobile/Auth/PennAuthRequestable.swift
```diff
@@ -43,7 +43,8 @@ extension PennAuthRequestable {
                 }
             } else {
                 UserDefaults.standard.setShibbolethAuth(authedIn: false)
-                completionHandler(nil, nil, NetworkingError.authenticationError)
+//                completionHandler(nil, nil, NetworkingError.authenticationError)
+                completionHandler(nil, nil, NetworkingError.noInternet)
```
joshdoman: This isn't an accurate way of checking for no internet. Let's use [this example](https://github.com/pennlabs/penn-mobile-ios/blob/master/PennMobile/Home/Networking/HomeAPIService.swift#L40) to check for no internet at the very beginning of the datatask callback. The issue with failing with no internet here is that it may fail for a variety of reasons, such as the server not returning a valid HTTP response or the data not decodable as a string. Instead of passing back a `noInternet` error though, let's pass back an `other` error here.

PennMobile/Auth/PennAuthRequestable.swift
```diff
@@ -43,8 +49,7 @@ extension PennAuthRequestable {
                 }
             } else {
                 UserDefaults.standard.setShibbolethAuth(authedIn: false)
-//                completionHandler(nil, nil, NetworkingError.authenticationError)
-                completionHandler(nil, nil, NetworkingError.noInternet)
```
01jongmin: You mentioned to me in a previous comment to change this completionHandler to return a NetworkingError.other.
I just wanted to note that it was originally a .authenticationError before I commented it out.

PennMobile/Auth/PennLoginController.swift
```diff
@@ -147,9 +147,11 @@ class PennLoginController: UIViewController, WKUIDelegate, WKNavigationDelegate,
         }
     }
     
+    var handleCancel: (() -> Void)? = nil
```
01jongmin: I added a variable in which we can handle the case in which the user decides to simply cancel the login screen.
For my PAC Code vc, I thought it would be ideal if we just make the user go back to the "More" screen if they decide not to login when they should.

PennMobile/New Group/PAC Code/PacCodeViewController.swift
```diff
@@ -58,22 +61,59 @@ class PacCodeViewController : UIViewController, ShowsAlert, IndicatorEnabled {
...
+        pacCodeTitleLabel.text = "PAC Code"
+        
+        view.addSubview(pacCodeTitleLabel)
+        pacCodeTitleLabel.translatesAutoresizingMaskIntoConstraints = false
+        pacCodeTitleLabel.bottomAnchor.constraint(equalTo: pacCodeSecurityInfoLabel.topAnchor, constant: -20).isActive = true
+        pacCodeTitleLabel.centerXAnchor.constraint(equalTo: view.safeAreaLayoutGuide.centerXAnchor).isActive = true
+    }
+    
+//    This is only in place as a placeholder for an icon if deemed necessary
+    func setupPlaceHolderIcon() {
```
01jongmin: The UI has changed as shown below.
Simulator Screen Shot - iPhone 11 Pro Max - 2020-03-09 at 18 22 36

I thought it may be a good idea to have a sort of icon at the top of the vc to avoid making the page look so dull. I'm however completely down from removing it if you feel simplicity should be a priority.

PennMobile/New Group/PAC Code/PacCodeViewController.swift
```diff
@@ -103,38 +143,32 @@ extension PacCodeViewController : KeychainAccessible, LocallyAuthenticatable {
...
             self.pacCode = pacCode
+            updatePACCode()
         } else {
             if Account.isLoggedIn {
-                self.showActivity()
-                
                 // Handle the case in which the user is logged in but hasn't yet fetched their PAC Codes
-                // Acts as though the user pressed the refresh button
-                PacCodeNetworkManager.instance.getPacCode { result in self.handleNetworkPacCodeRefreshCompletion(result) }
+                handleNetworkPacCodeRefetch()
```
01jongmin: The code that dealt with showing the activity indicator, refetching of the PAC code using the network manager, handling the response, and finally hiding the indicator has been refactored away into a new function called `handleNetworkPacCodeRefetch()`  in line 180 -185.

01jongmin:
Below are video links of the changes
- [Success](https://youtu.be/4dTqeRunMFk)
- [Login Error](https://youtu.be/B4I383fioCk)
- [Authentication](https://youtu.be/203KYHFC7O8)

PennMobile/New Group/PAC Code/PacCodeViewController.swift
joshdoman: I like the graphic!

PR #318, opened 2020-03-10
Bug fix for dining favorite

01jongmin:


01jongmin:
It seems to be the case that the `HomeDiningCellItem` had to variables that needed to be updated: `venues` and `venuesIds`

However, the 'venuesIds' array was not being updated.

The merge conflict has to do with an error from my prior commit. I made the mistake of not staging the changes made to add the CourseScheduleViewController and its model into the directory of the project. I apologize for any trouble caused by this mistake.

PennMobile/Home/Controllers/HomeViewController.swift
```diff
@@ -262,6 +262,7 @@ extension HomeViewController : DiningCellSettingsDelegate {
             diningItem.venues = DiningDataStore.shared.getVenues(for: DiningVenue.defaultVenueIds)
         } else {
             diningItem.venues = DiningDataStore.shared.getVenues(for: venueIds)
+            diningItem.venueIds = venueIds
```
joshdoman: This makes sense but let's do the same for the other side of the if statement if the selected venue ids array is empty. In that case, we want to display the default venue ids.

PR #320, opened 2020-03-13
Error handling for CourseSchedule VC

01jongmin:
Changed return type for `getCourses` function of `PennInTouchNetworkManager` to `Result`

Added error handling for `courseSchedule` View Controller

Videos:
[Refresh Controller](https://youtu.be/MYdAYmIyiGQ) 
[Login Logout remove course schedule](https://youtu.be/D4r11miZ-KQ)

### DIFF
--- a/PennMobile/Courses/CourseScheduleViewController.swift
+++ b/PennMobile/Courses/CourseScheduleViewController.swift
@@ -9,7 +9,7 @@ import UIKit
 
 protocol CourseScheduleCellDelegate: ModularTableViewCellDelegate, BuildingMapSelectable {}
 
-class CourseScheduleViewController: GenericViewController, IndicatorEnabled, ShowsAlert {
+class CourseScheduleViewController: GenericViewController, IndicatorEnabled, ShowsAlertForError {
 
     fileprivate var courseScheduleTableView: ModularTableView!
     fileprivate var model: CourseScheduleTableViewModel!
@@ -62,7 +62,6 @@ extension CourseScheduleViewController {
         })
     }
 
-    // TODO: Need review: ad-hoc method to overcome conflict of UIAlertController and UIRefreshController
     override func viewDidDisappear(_ animated: Bool) {
         super.viewDidDisappear(animated)
         self.courseScheduleTableView.isHidden = true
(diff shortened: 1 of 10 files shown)


PennMobile/Courses/CourseScheduleViewController.swift
```diff
@@ -56,18 +34,143 @@ extension CourseScheduleViewController {
...
+            }
+        })
+    }
+
+    // TODO: Need review: ad-hoc method to overcome conflict of UIAlertController and UIRefreshController
+    override func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
+        self.courseScheduleTableView.isHidden = true
+        self.courseScheduleTableView.refreshControl?.endRefreshing()
+        self.courseScheduleTableView.contentOffset = CGPoint.zero
```
01jongmin: This is an ad-hoc method.
For some reason, UIAlertController and UIRefreshController seems to be in conflict with one another that prevents me from seeing the effect of calling `endRefreshing()`

Similar issues has been brought up by other [people](https://stackoverflow.com/questions/40230892/uirefresh-control-endrefreshing-doesnt-work). I wasn't sure if you knew of a better work around way of doing this.

PennMobile/Courses/CourseScheduleViewController.swift
```diff
@@ -56,18 +34,143 @@ extension CourseScheduleViewController {
...
+            self.showAlert(withMsg: "You appear to be offline.\nPlease try again later.", title: "Network Error", completion: { self.navigationController?.popViewController(animated: true) } )
+
+        case .failure(.parsingError):
+            self.showAlert(withMsg: "Penn's course servers are currently not updating. We hope this will be fixed shortly", title: "Uh oh!", completion: { self.navigationController?.popViewController(animated: true) })
+
+        case .failure(.authenticationError):
+            self.showAlert(withMsg: "Unable to access your courses.\nPlease login again.", title: "Login Error", completion: { self.handleAuthentication() })
+
+        default:
+            self.showAlert(withMsg: "Something went wrong.\nPlease try again later.", title: "Uh oh!", completion: { self.courseScheduleTableView.refreshControl?.endRefreshing(); self.navigationController?.popViewController(animated: true) } )
```
joshdoman: Let's look into abstracting this error handling into a protocol, perhaps one that inherits from ShowsAlert (and thus must be a view controller). We should avoid duplicate code, especially when the error messages are the same.

01jongmin: Got it. I'll work on it right away!

PennMobile/General/Protocols.swift
```diff
@@ -121,6 +121,37 @@ extension ShowsAlert where Self: UIViewController {
...
+            self.showAlert(withMsg: "You appear to be offline.\nPlease try again later.", title: "Network Error", completion: noInternet)
+
+        case .failure(.parsingError):
+            self.showAlert(withMsg: "Penn's \(title) servers are currently not updating. We hope this will be fixed shortly", title: "Uh oh!", completion: parsingError)
+            
+        case .failure(.serverError):
+            self.showAlert(withMsg: "Penn's \(title) servers are currently not updating. We hope this will be fixed shortly", title: "Uh oh!", completion: serverError)
+            
+        case .failure(.jsonError):
+            self.showAlert(withMsg: "Penn's \(title) servers are currently not updating. We hope this will be fixed shortly", title: "Uh oh!", completion: jsonError)
```
01jongmin: I wasn't too sure if it was the correct decision to make the messaging for internet, parsing, and json error all the same, informing the users that the issue wasn't on our end.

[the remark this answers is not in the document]
r3khaan: Now that I think about it, there's also a strong case for just changing the name. Adding extra functionality for customizing the success message & title is something that would only be useful later (when we have more than 1 type of success message). We can always change it then.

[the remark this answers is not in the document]
01jongmin: What about we keep the protocol name as `ShowsAlertForError` and change the function name to `showRefreshAlertForError`. We can then add additional functions that respond to error for different type of actions later down the line.

PR #335, opened 2020-08-19
Cleared up project warnings

r3khaan:
- Updated ImageNetworkManager to use results instead of standard completion handlers (old way is now deprecated)
- Resolved two other minor warnings
- Getting back in the practice of using PR's!


01jongmin (approved):
LGTM!

Issue #336, opened 2020-08-19
Date Parsing Issue


r3khaan:
Calendar on home page is showing events off by 1 day. Probably a timezone parsing issue.

PR #337, opened 2020-08-19
Date parsing bug

r3khaan:
Attempts to fix issue #336.

Not sure if this is the best way to do it, any help is appreciated.

I basically just set the time to 6am EST, when parsing dates from https://api.pennlabs.org/calendar/.

### DIFF
--- a/PennMobile/Home/Cells/Calendar/CalendarAPI.swift
+++ b/PennMobile/Home/Cells/Calendar/CalendarAPI.swift
@@ -30,19 +30,21 @@ class CalendarAPI: Requestable {
                 var events = [CalendarEvent]()
                 for json in jsonArray {
                     let name = json["name"].stringValue
-                    let start = json["start"].stringValue
-                    let end = json["end"].stringValue
+
+                    //set time to 6 am EST
+                    let start = "\(json["start"].stringValue) 6:00 AM"
+                    let end = "\(json["end"].stringValue) 6:00 AM"
                     
                     let dateFormatter = DateFormatter()
-                    dateFormatter.dateFormat = "yyyy-MM-dd"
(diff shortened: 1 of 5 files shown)


PennMobile/Home/Cells/Calendar/CalendarAPI.swift
```diff
@@ -30,19 +30,21 @@ class CalendarAPI: Requestable {
...
+
+                    //set time to 6 am EST
+                    let start = "\(json["start"].stringValue) 6:00 AM"
+                    let end = "\(json["end"].stringValue) 6:00 AM"
                     
                     let dateFormatter = DateFormatter()
-                    dateFormatter.dateFormat = "yyyy-MM-dd"
+                    dateFormatter.dateFormat = "yyyy-MM-dd h:mm a"
                     dateFormatter.timeZone = TimeZone(abbreviation: "EST")
+
```
01jongmin, line 41: I'm not exactly getting why the addition/removal of the hour portion of the time string is causing/fixing the bug.

r3khaan: @01jongmin Yeah, I think the problem is that when it parses the date from https://api.pennlabs.org/calendar/, it sets the time to 12:00 AM (since the API does not specify the time). And 12:00 am on Aug 24 in EST = 9:00 pm on Aug 23 in PST (and in other timezones too). So, if I manually set the time to 6:00 am instead of 12:00 am, then it will show aug 24 in more timezones.

Still feels like a bit of a hack, but seems to work for almost all timezones. If you have other ideas, shoot.

01jongmin: Oh got it. That makes sense. It does feel a bit like a hack, but I don't think it will lead to any breaking problems down in the future for us. LGTM!

PR #340, opened 2020-09-02
Update about page

r3khaan:
- Moved old members to alumni section, added new members
- Now sorting members alphabetically instead of arbitrarily

### DIFF
--- a/PennMobile.xcodeproj/project.pbxproj
+++ b/PennMobile.xcodeproj/project.pbxproj
@@ -2125,7 +2125,7 @@
 				CODE_SIGN_ENTITLEMENTS = PennMobile/PennMobile.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				CODE_SIGN_STYLE = Automatic;
-				CURRENT_PROJECT_VERSION = 6511;
+				CURRENT_PROJECT_VERSION = 6520;
 				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
 				DEVELOPMENT_TEAM = VU59R57FGM;
 				GCC_PREFIX_HEADER = PennMobile/Supporting_Files/PrefixHeader.pch;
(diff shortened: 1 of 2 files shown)


PennMobile.xcodeproj/project.pbxproj
```diff
@@ -2125,7 +2125,7 @@
 				CODE_SIGN_ENTITLEMENTS = PennMobile/PennMobile.entitlements;
 				CODE_SIGN_IDENTITY = "iPhone Developer";
 				CODE_SIGN_STYLE = Automatic;
-				CURRENT_PROJECT_VERSION = 6511;
+				CURRENT_PROJECT_VERSION = 6520;
```
01jongmin, line 2128: I'm not too familiar with the project.pbxproj file but do you know what caused this change as well as the ones below?

PennMobile/About/Controllers/AboutViewController.swift
```diff
@@ -22,6 +23,21 @@ class Member {
...
+            if lhsLast != rhsLast {
+                return lhsLast < rhsLast
+            }
+        }
+        return lhs.name < rhs.name
+    }
+
+    static func == (lhs: Member, rhs: Member) -> Bool {
+        return lhs.name == rhs.name && lhs.imageURL == rhs.imageURL
+    }
```
01jongmin, line 40: This isn't a huge issue, but my name, for example, has two spaces so writing a function like this is not the desired behavior (ie. my last variable will be "Min Choi" not "Choi")

Rather than splitting the name component, it might be better to replace the name variable with two variables: firstName and lastName. I think this would increase readability as you no longer have to include the components(seperatedBy: ) in the comparison.

r3khaan: Good point!

PennMobile.xcodeproj/project.pbxproj
r3khaan: I actually meant to update this in a separate branch, but got lazy / didn't want to forget. Either way, it's probably better to make this version change at the very end.

01jongmin:
LGTM!

01jongmin (approved):
LGTM!

PR #345, opened 2020-09-06
Alumni name fix

01jongmin:
Added check to see if the first name is an empty string and if so, just say Howdy!

### DIFF
--- a/PennMobile/Home/Controllers/HomeViewController.swift
+++ b/PennMobile/Home/Controllers/HomeViewController.swift
@@ -68,7 +68,7 @@ class HomeViewController: GenericViewController {
             let firstName = Account.getAccount()?.first ?? GSRUser.getUser()?.firstName
             if let firstName = firstName {
                 let intros = ["Welcome", "Howdy", "Hi there", "Hello", "Greetings", "Sup"]
-//                self.displayTitle = "\(intros.random!), \(firstName)!"
+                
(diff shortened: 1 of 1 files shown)


PennMobile/Home/Controllers/HomeViewController.swift
```diff
@@ -68,7 +68,13 @@ class HomeViewController: GenericViewController {
...
+                
+                if firstName != "" {
+                    self.displayTitle = "\(intros.random!), \(firstName)!"
+                } else {
+                    self.displayTitle = "Howdy!"
+                }
+                
                 titleCacheTimestamp = Date()
             } else {
                 self.displayTitle = "Home"
```
r3khaan, line 80: I think we should display the same message for users, who are not signed in, & alumni. Both should say "home" or both should give a customized greeting like "howdy." 

I like the latter option more, what do you think?

r3khaan (changes requested):
Sweet and simple! Before merging in the PR though, I think we should make the experience consistent across users that are not logged in and alumni (see comment above)

r3khaan:
@01jongmin

01jongmin:
Apologize for the delay. The code should now follow the later option.

r3khaan (changes requested):
Very minor change! No need to re-request my review after, feel free to squash and merge yourself.

PR #347, opened 2020-09-27
Bug fixes for xcode 12

r3khaan:
- Removed WKZombie to make the project runnable on xcode 12. 
- Obviously breaks 2FA (but it was already broken)


r3khaan:
Don't delete this branch (for when we need to potentially add WKZombie back)

PR #350, opened 2020-10-07
GSRs Unavailable 

raunaqsingh2020:
Changed GSRController to show closed view when there are 0 rooms and to show the empty view (no results found) only when there are rooms but none available for the specific times.

E.g. GSR with rooms: 
Screen Shot 2020-10-07 at 6 12 21 PM

E.g. GSR with rooms but none available at time given by slider:
Screen Shot 2020-10-07 at 6 13 17 PM

E.g. GSR with 0 rooms:
Screen Shot 2020-10-07 at 6 12 32 PM


PennMobile/GSR-Booking/Views/ClosedView.swift
```diff
@@ -0,0 +1,52 @@
...
+        l.text = "This GSR is closed until further notice."
+        l.textColor = UIColor.labelSecondary
+        l.textAlignment = .center
+        l.numberOfLines = 0
+        l.translatesAutoresizingMaskIntoConstraints = false
+        return l
+    }()
+    
+    private let closedImage: UIImageView = {
+        let imgView = UIImageView(image: UIImage(named: "EmptyStateGSR")!)
```
r3khaan, line 25: Not a big deal, but I don't think we need to force unwrap the `UIImage` here?

PennMobile/GSR-Booking/Views/ClosedView.swift
```diff
@@ -0,0 +1,52 @@
...
+        addSubview(closedImage)
+        closedImage.widthAnchor.constraint(equalToConstant: 200).isActive = true
+        closedImage.centerYAnchor.constraint(equalTo: centerYAnchor, constant: -100).isActive = true
+        closedImage.centerXAnchor.constraint(equalTo: centerXAnchor).isActive = true
+        
+        addSubview(label)
+        
+        label.centerYAnchor.constraint(equalTo: centerYAnchor, constant: 75).isActive = true
+        label.centerXAnchor.constraint(equalTo: centerXAnchor).isActive = true
+        label.widthAnchor.constraint(equalTo: widthAnchor, constant: -10).isActive = true
```
r3khaan, line 45: Nice!!

PennMobile/GSR-Booking/Views/ClosedView.swift
```diff
@@ -0,0 +1,52 @@
+//
+//  ClosedView.swift
```
r3khaan, line 2: Let's call this file/class something more specific, like `GSRClosedView`

PennMobile/GSR-Booking/Controllers/GSRController.swift
```diff
@@ -146,12 +157,16 @@ extension GSRController: GSRViewModelDelegate {
...
+            
         }
+        
+        
     }
 
     func refreshDataUI() {
-        emptyView.isHidden = !viewModel.isEmpty
         tableView.isHidden = viewModel.isEmpty
+        closedView.isHidden = viewModel.existsNonEmptyRoom()
```
r3khaan, line 168: Ahh .. I totally see why `existsNonEmptyRoom` is not good enough for our purposes here. One way to resolve it:

Instead of checking whether at least one `room` exists, we should check if at least one `timeSlot` exists! If 0 `timeSlot`'s exists, then we should show the `closedView`. Otherwise, we should hide it.

r3khaan (changes requested):
Great start, Raunaq! I'm glad you were able to figure out the programmatic UI. I added some small comments here and there, but here are the major points:

Ideally, we want to show the same "closed" message + image for each GSR location (because all of them are actually closed right now). I should have explained that better, because I'm sure you're still a little confused about what a GSR location versus room versus timeslot is. 
- I added a suggestion for how to handle the case where a GSR location has >=1 rooms, but has 0 timeslots in the comments. 
- Try to see if you can figure out how to handle the case where the server doesn't return a valid response (e.g. for Huntsman). Hint: go to the `fetchData` function in the `GSRController.swift` file and see if you can figure it out.

PennMobile/GSR-Booking/Controllers/GSRController.swift
jtlieb: Just want to add something here -- I think we may want to add a different empty state rather than "closed" for when there are no empty rooms available. Def agree with Rehaan here, "Closed" should be used for an empty schedule for the building, which would signify that it's being closed for a significant amount of time"

jtlieb:
From a layout perspective, I think the illustration needs to be brought down closer to the center of the whitespace, currently it feels a little bit cramped below the slider bar.

raunaqsingh2020:
Fixed the layout for the closed view by lowering the sign a bit. Fixed when this view shows:

When the app cannot get a valid response from the server (ex. Huntsman):

Screen Shot 2020-10-11 at 7 31 25 PM

When the GSR has 0 `timeslots`:

Screen Shot 2020-10-11 at 7 31 52 PM

GSR with 1+ `timeslots` (added these manually to test):

Screen Shot 2020-10-11 at 7 32 05 PM

GSR with 1+ `timeslots` when the slider does not include the times:

Screen Shot 2020-10-11 at 7 32 14 PM

Let me know if there are any other changes to make. Thanks!

r3khaan:
@raunaqsingh2020 Hey, will take a look soon

PennMobile/GSR-Booking/ViewModel/GSRViewModel.swift
```diff
@@ -202,6 +202,11 @@ extension GSRViewModel: GSRSelectionDelegate {
     func clearSelection() {
         currentSelection = []
     }
+    
+    func existsTimeSlot() -> Bool {
+        let roomsWithTimeSlots = allRooms.filter { $0.timeSlots.count > 0 }
+        return roomsWithTimeSlots.count > 0
+    }
```
r3khaan, line 209: this is some clean code 😎

PennMobile/GSR-Booking/Controllers/GSRController.swift
```diff
@@ -134,24 +145,31 @@ extension GSRController: GSRViewModelDelegate {
...
                     self.refreshDataUI()
                     self.rangeSlider.reload()
                     self.refreshBarButton()
                     self.stopLoadingViewAnimation()
+                }else{//no valid response
+                    self.viewModel.updateData(with: [])
+                    self.refreshDataUI()
+                    self.rangeSlider.reload()
+                    self.refreshBarButton()
+                    self.stopLoadingViewAnimation()
```
r3khaan, line 163: Suggestion for readability: You can remove the `if ... else ...` and instead write `let rooms = rooms ?? []` at the start. This removes the duplicate code.

r3khaan (changes requested):
Added a minor formatting suggestion for code readability. Thanks for providing screenshots of the different cases, and testing for when there actually are timeslots.

r3khaan (approved):
LGTM!

PR #358, opened 2020-10-22
Homescreen non-interactive bugfix

01jongmin:
The issue was that UITableViewCells have a contentView. This was placed in front of the various views in our own custom table view cells and blocked touches to any views behind it. 

Setting the isUserInteractionEnabled of contentView to false has allowed us to re-enable the various touch detection that was necessary for our homescreen.


r3khaan (approved):
I'm feeling both happy and sad that it only took 1 line to fix this bug .. good find regardless!

PR #381, opened 2021-03-17
Update MoreViewController.swift

bsk42:
Added link to air table for user feedback


r3khaan (changes requested):
Sweet, the only change I'd make is instead of calling the tab "User Feedback," I'd call it "Share Your Feedback"

PR #397, opened 2021-10-03
Profile Page Completed

aantenberg:
simulator_screenshot_624D7D2D-6EAC-428E-97D5-921586216F4E


PennMobile/New Group/AccountPageViewController.swift
```diff
@@ -0,0 +1,129 @@
...
+    var educationInfo = [(text: "Graduation Term", info: " "), (text: "School", info: " "), (text: "Major", info: " ")]
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        setupView()
+        setupTableView()
+        guard Account.isLoggedIn else {
+            self.showAlert(withMsg: "Please login to use this feature", title: "Login Error", completion: { self.navigationController?.popViewController(animated: true)} )
+            return
+        }
```
01jongmin, line 26: You can use if let, and set a root level Account object to avoid the unwrapping in subsequent calls.

PennMobile/New Group/AccountPageViewController.swift
```diff
@@ -0,0 +1,129 @@
...
+        
+        if schoolsSet.count > 1 {
+            educationInfo[1].text += "s"
+        }
+        
+        if majorsSet.count > 1 {
+            educationInfo[2].text += "s"
+        }
+        educationInfo[1].info = Array(schoolsSet).joined(separator: ", ")
```
01jongmin, line 82: You can use ListFormatter.localizedString(byJoining:)

01jongmin (changes requested):
Looks great! A couple of changes requested in the comments!

PennMobile/New Group/AccountPageViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+//
+
+import Foundation
+import UIKit
+
+class AccountPageViewController: UIViewController, ShowsAlertForError, UITableViewDelegate, UITableViewDataSource {
+    var account: Account!
+    let tableView = UITableView(frame: .zero, style: .insetGrouped)
+    
+    var profileInfo = [(text: "Name", info: " "), (text: "Username", info: " "), (text: "Email", info: " ")]
```
01jongmin, line 16: Rather than initializing them with missing values, it would be ideal for you to build them up in the method below by appending them. 

`var profileInfo: [(text: String, info: String)] = []`

PennMobile/New Group/AccountPageViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+        tableView.estimatedRowHeight = 600
+        tableView.delegate = self
+        tableView.dataSource = self
+    }
+    
+    func setupProfileInfo() {
+        guard let firstName = account.first, let lastName = account.last else {
+            return
+        }
+        profileInfo[0].info = "\(firstName) \(lastName)"
```
01jongmin, line 55: It would be easier for the reader if you instead did
`profileInfo.append((text: "name", info: "\(firstName) \(lastName)"))`
so the value of the text and info are in the same place. If we need to quickly add a new entry, this way we don't need to edit two separate locations in the file.

PennMobile/New Group/AccountPageViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+
+import Foundation
+import UIKit
+
+class AccountPageViewController: UIViewController, ShowsAlertForError, UITableViewDelegate, UITableViewDataSource {
+    var account: Account!
+    let tableView = UITableView(frame: .zero, style: .insetGrouped)
+    
+    var profileInfo = [(text: "Name", info: " "), (text: "Username", info: " "), (text: "Email", info: " ")]
+    var educationInfo = [(text: "Graduation Term", info: " "), (text: "School", info: " "), (text: "Major", info: " ")]
```
01jongmin, line 17: Same for educationInfo

01jongmin (changes requested):
Just a couple of more minor edits!

01jongmin (approved):
LGTM

PR #418, opened 2022-02-25
Launch alert

kunlizhang:
New splash screen with PennMobile icon, alerts user if version is outdated.

### DIFF
--- a/PennMobile/Setup + Navigation/RootViewController.swift
+++ b/PennMobile/Setup + Navigation/RootViewController.swift
@@ -12,7 +12,7 @@ import StoreKit
 import SwiftyJSON
 
 // Source: https://medium.com/@stasost/ios-root-controller-navigation-3625eedbbff
-class RootViewController: UIViewController, NotificationRequestable {
+class RootViewController: UIViewController, NotificationRequestable, ShowsAlert {
     var current: UIViewController
 
     private var lastLoginAttempt: Date?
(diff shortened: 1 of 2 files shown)


01jongmin (changes requested):
You shouldn't be modifying any other files other than the RootViewController and SplashViewController. Run `git reset HEAD~` and restore the changes in the other two files. Lastly, run the process of adding, commiting, then run `git push --force` instead of simply `git push` to overwrite this history.

PennMobile/Setup + Navigation/RootViewController.swift
```diff
@@ -39,18 +39,38 @@ class RootViewController: UIViewController, NotificationRequestable {
 
         if #available(iOS 15.0, *) {
             Task {
-                if let (data, _) = try? await URLSession.shared.data(from: URL(string: "https://itunes.apple.com/lookup?bundleId=org.pennlabs.PennMobile")!),
+                if let (data, _) = try? await URLSession.shared.data(from: URL(string: "https://itunesl.apple.com/lookup?bundleId=org.pennlabs.PennMobile")!),
```
01jongmin, line 42: This should be itunes not itunesl I believe.

01jongmin (changes requested):
Just one more change!

01jongmin (approved):
LGTM, feel free to merge it in 👍

PR #425, opened 2022-09-12
Allow for empty captions/authors

kunlizhang:
Native news was crashing because API would sometimes not provide captions for images, and the code originally assumed that authors was non-empty.


01jongmin (approved):
LGTM! Merge it in and we can release it as 6.8.0

PR #426, opened 2022-10-01
fix notification settings page

raunaqsingh2020:
Hit endpoint at `https://pennmobile.org/api/user/notifications/settings/` in order to get user's notification settings, and replace `NotificationsTableViewController` with a SwiftUI View.

NOTE: Should remove previous code (`NotificationPreference.swift`, `NotificationsTableViewController.swift`, `NotificationsTableViewCell.swift`, and unused/outdated methods in `UserDBManager.swift` and `UserDefaults + Helpers.swift`) when there is an endpoint for privacy settings. Previously, notification and privacy settings were coupled together within the same route.


01jongmin (approved):
LGTM

PR #429, opened 2022-10-05
Fixed dining analytics bugs (now predicts from maximum balance)

JHawk0224:
Summary of changes:
- The red dot is centered
- The dates in the graph’s text is not cut off
- Dollar predictions now says “$xxx.00” instead of “xxx.00 Dollars” (so text isn’t cut off)
- The graph uses the (first) maximum value as the start point for predictions
- Rewrote some of the graph stuff (specifically PredictionSlopePath and everything that made a call to it) to take in a slope instead of predictionZeroPoint so we can make the slope as close to 0 as we want.
- In DiningAnalyticsViewModel, condensed the prediction functions into one function that returns the slope of the line, the predicted zero date, and the predicted end balance.
- Slope of infinity (just got dining plan and have/haven’t spent stuff) and slope of 0 (haven’t spent anything on new dining plan) are handled. Both draw a horizontal prediction line.

### DIFF
--- a/PennMobile/Dining/SwiftUI/DiningAnalyticsViewModel.swift
+++ b/PennMobile/Dining/SwiftUI/DiningAnalyticsViewModel.swift
@@ -23,12 +23,13 @@ class DiningAnalyticsViewModel: ObservableObject {
 
     @Published var dollarPredictedZeroDate: Date = Date.endOfSemester
     @Published var predictedDollarSemesterEndBalance: Double = 0
-    @Published var predictedDollarZeroPoint: PredictionsGraphView.YXDataPoint = PredictionsGraphView.YXDataPoint(y: 0, x: 0)
     @Published var swipesPredictedZeroDate: Date = Date.endOfSemester
     @Published var predictedSwipesSemesterEndBalance: Double = 0
-    @Published var predictedSwipesZeroPoint: PredictionsGraphView.YXDataPoint = PredictionsGraphView.YXDataPoint(y: 0, x: 0)
     @Published var swipeAxisLabel: ([String], [String]) = ([], [])
     @Published var dollarAxisLabel: ([String], [String]) = ([], [])
+    @Published var dollarSlope: Double = 0.0
+    @Published var swipeSlope: Double = 0.0
+    
     var yIntercept = 0.0
     var slope = 0.0
     let formatter = DateFormatter()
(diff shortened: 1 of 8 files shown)


PennMobile/Dining/SwiftUI/DiningAnalyticsViewModel.swift
```diff
@@ -67,36 +68,46 @@ class DiningAnalyticsViewModel: ObservableObject {
...
-    func predictSemesterEndBalance(firstBalance: DiningAnalyticsBalance, lastBalance: DiningAnalyticsBalance) -> Double {
-        // y = slope * x + yIntercept
-        let yIntercept = firstBalance.balance
-        let slope = (lastBalance.balance - firstBalance.balance) / Double(Calendar.current.dateComponents([.day], from: firstBalance.date, to: lastBalance.date).day!)
-        let diffInDays = Calendar.current.dateComponents([.day], from: firstBalance.date, to: Date.endOfSemester).day!
-        return ((slope * Double(diffInDays)) + yIntercept)
+    func getPredictions(firstBalance: DiningAnalyticsBalance, lastBalance: DiningAnalyticsBalance) -> (slope: Double, predictedZeroDate: Date, predictedEndBalance: Double) {
+        if firstBalance.date == lastBalance.date {
+            let zeroDate = Calendar.current.date(byAdding: .day, value: 1, to: Date.endOfSemester)!
+            return (Double.infinity, zeroDate, lastBalance.balance)
```
aantenberg, line 94: Curious why we're returning slope as infinity in this case. Maybe mathematically it is more accurate, but it seems to be treated the same as 0 in PredictionSlopePath, and I think in the context of predictions it makes sense for 0 to be the max slope. For that reason, I think you can get rid of this `if` statement altogether, as it should get caught in the `firstBalance.balance == lastBalance.balance` case. Please let me know if you agree!

PennMobile/Dining/SwiftUI/Views/Insights/PredictionsGraph/VariableStepLineGraphView+PredictionSlopePath.swift
```diff
@@ -27,16 +27,26 @@ extension VariableStepLineGraphView {
...
                 x: lastDataPoint.x * rect.maxX,
                 y: rect.maxY - (rect.maxY * lastDataPoint.y)
             ))
-
-            path.addLine(to: CGPoint(
-                x: predictionZeroPoint.x * rect.maxX,
-                y: rect.maxY - (rect.maxY * predictionZeroPoint.y)
-            ))
+            if slope == 0 {
+                path.addLine(to: CGPoint(
```
aantenberg, line 35: Why does this have to be its own case? What is the behavior if slope is 0 and we use the regular `x: rect.maxX, y: rect.maxY * (1 - lastDataPoint.y - slope * (1 - lastDataPoint.x))` case?

PennMobile/Dining/SwiftUI/Views/Insights/PredictionsGraph/PredictionsGraphView.swift
```diff
@@ -79,7 +79,7 @@ struct PredictionsGraphView: View {
                     // "Leftover" Dollars, wasted dollars
                     Text(displayZeroDate ? ("Out of \(balanceType == .swipes ? "Swipes" : "Dollars")") : "Extra Balance")
                         .font(.caption)
-                    Text(displayZeroDate ? "\(self.formattedZeroDate)" : "\(formattedBalance)\(balanceType == .swipes ? " Swipes" : " Dollars")")
+                    Text(displayZeroDate ? "\(self.formattedZeroDate)" : (balanceType == .swipes ? "\(formattedBalance) Swipes" : "$\(formattedBalance)"))
```
aantenberg, line 82: Nice! Not sure why we didn't do this in the first place.

PennMobile/Dining/SwiftUI/DiningAnalyticsViewModel.swift
```diff
@@ -67,36 +68,46 @@ class DiningAnalyticsViewModel: ObservableObject {
...
+        } else if firstBalance.balance == lastBalance.balance {
+            let zeroDate = Calendar.current.date(byAdding: .day, value: 1, to: Date.endOfSemester)!
+            return (Double(0.0), zeroDate, lastBalance.balance)
+        } else {
+            let yIntercept = firstBalance.balance
```
aantenberg, line 99: I like that you combined this all into one function to avoid repeated code, but it seems like the repeated code from the old `predictZeroDate` and `predictSemesterEndBalance` functions were just calculating the slope. I think this might be a bit more readable if we kept `predictZeroDate`, `predictSemesterEndBalance`, and write a `getSlope` function. That way, `predictZeroDate` and `predictSemesterEndBalance` can either have slope passed in as a parameter or just call `getSlope` within their own functions. Then,`getPredictions` could look like this:

```
func getPredictions(firstBalance: DiningAnalyticsBalance, lastBalance: DiningAnalyticsBalance) -> (slope: Double, predictedZeroDate: Date, predictedEndBalance: Double) {
  if ____ {
    ...
  }
  var slope = getSlope(firstBalance, lastBalance)
  let zeroDate = predictZeroDate(firstBalance, lastBalance, slope)
  let endBalance = predictSemesterEndBalance(firstBalance, lastBalance, slope)
  return (slope, zeroDate, endBalance) 
}
```

I think it's just a little bit nicer to have the calculations abstracted out.

aantenberg (changes requested):
Really awesome job with this refactor and bug fix! The graph stuff is really difficult to work with, so I'm impressed with the slope refactor. I left a few questions, so take a look whenever is convenient!

PennMobile/Dining/SwiftUI/DiningAnalyticsViewModel.swift
JHawk0224: Yea I think that makes sense. I'll combine these two cases together and just return 0.0.

PennMobile/Dining/SwiftUI/DiningAnalyticsViewModel.swift
JHawk0224: I can definitely separate out the slope calculation and the others so it's more readable. However, I would like to note that currently, the slope used to calculate the zeroDate and the endBalance is different from the slope that's returned from getPredictions. The prior one is needed to calculate the zeroDate, but the returned one is the one that's used for the graph (it's scaled differently). I'm going to look into the exact differences between them and see if I can simplify it before fixing it.

JHawk0224: I don't really see any way to get the correct slope for the graph from the original slope value (for calculating zeroDate) without already doing what is done, so I left it at that.

JHawk0224:
I also just pushed a changed where all of the dining halls have links to their appropriate menus.

aantenberg (changes requested):
Just one more small thing, but otherwise LGTM!

aantenberg (approved):
Looks good! Ready to be merged!!

PR #431, opened 2022-10-14
Update mobile backend notifications endpoints

raunaqsingh2020:
Use `PATCH` request with service `id` instead of `POST` request with service name / title to update user notification preferences.


01jongmin (approved):
LGTM!

PR #432, opened 2022-10-14
Course Schedule: Show login controller on authentication failure

anli5005:
Sometimes Path@Penn doesn't give us a token if we attempt a login, likely due to some sort of expired cookie. This PR implements a quick and dirty SwiftUI `LabsLoginView` and presents it if it detects an authentication failure.


anli5005:
Shoot, it seems like my `swiftlint --fix` invocation touched a few dining files. Let me know if I should revert those.

01jongmin (approved):
LGTM!

anli5005:
Well, here goes nothing!

PR #433, opened 2022-10-16
Make dining balance calls async, add no dining plan found page

aantenberg:
Changed dining balance API calls to async, and added dining analytics "no dining plan found" page when a user opens the dining analytics page but do not have a dining plan. Also handling the case where a user has a swipe plan, but no dollar plan, and vice versa, by only showing the graph that has data. (Note: I'm not sure if this is actually possible by the University's meal plan options, but it can't hurt to handle it.)

simulator_screenshot_D584A9A8-511C-43B8-AA84-61EF03F5A270


PennMobile/Dining/SwiftUI/Views/Venue/DiningVenueView.swift
```diff
@@ -61,7 +59,7 @@ struct CustomHeader: View {
     }
 
     var body: some View {
-        HStack {
+        ZStack {
```
JHawk0224, line 62: Shouldn't this stay as an HStack?

PennMobile/Dining/SwiftUI/Views/Venue/DiningVenueView.swift
```diff
@@ -73,7 +71,9 @@ struct CustomHeader: View {
                         showMissingDiningTokenAlert = true
                         return
                     }
-                    DiningViewModelSwiftUI.instance.refreshBalance()
+                    Task.init() {
+                        await DiningViewModelSwiftUI.instance.refreshBalance()
+                    }
```
01jongmin, line 76: You can add a `.task` modifier at the end of the view to trigger a async function. More details can be found [here](https://developer.apple.com/documentation/swiftui/view/task(priority:_:))

JHawk0224:
Looks good other than the comments above and the merge conflicts!

JHawk0224 (approved):
Looks good

PR #435, opened 2022-10-18
Made dining analytics graph use dining plan start date for prediction

JHawk0224:
Before, we updated dining analytics graph to use the maximum values, but I realized this doesn't always work. It's possible for a change in dining plan to reduce swipes or dollars, so this change is necessary.

Note, since I added a file, I wasn't sure if I should commit PennMobile.xcodeproj/project.pbxproj. Please let me know if I should do this.

### DIFF
--- a/PennMobile/Dining/Networking + Cache/DiningAPI.swift
+++ b/PennMobile/Dining/Networking + Cache/DiningAPI.swift
@@ -140,23 +140,22 @@ extension DiningAPI {
 
 // MARK: Current Dining Plan Start Date
 extension DiningAPI {
-    func getDiningPlanStartDate(diningToken: String, _ callback: @escaping (_ startDate: Date?) -> Void) {
+    func getDiningPlanStartDate(diningToken: String) async -> Result<Date, NetworkingError> {
         let url = URL(string: "https://prod.campusexpress.upenn.edu/api/v1/dining/currentPlan")!
         var request = URLRequest(url: url)
         request.httpMethod = "GET"
         request.setValue(diningToken, forHTTPHeaderField: "x-authorization")
-        let task = URLSession.shared.dataTask(with: request) { (data, response, _ ) in
-            if let httpResponse = response as? HTTPURLResponse, let data = data, httpResponse.statusCode == 200 {
-                let decoder = JSONDecoder()
-                let dateFormatter = DateFormatter()
-                dateFormatter.dateFormat = "yyyy-MM-dd"
-                decoder.dateDecodingStrategy = .formatted(dateFormatter)
-                let plan = try? decoder.decode(DiningPlan.self, from: data)
-                callback(plan?.start_date)
-                return
-            }
-            callback(nil)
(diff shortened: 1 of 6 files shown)


JHawk0224:
I think my PennMobile.xcodeproj/project.pbxproj is a little behind in version but it shouldn't matter.

PennMobile/Dining/Networking + Cache/DiningAPI.swift
```diff
@@ -137,3 +137,26 @@ extension DiningAPI {
         task.resume()
     }
 }
+
+// MARK: Current Dining Plan Start Date
+extension DiningAPI {
+    func getDiningPlanStartDate(diningToken: String, _ callback: @escaping (_ startDate: Date?) -> Void) {
```
aantenberg, line 143: If possible, can you make this async/await instead of a callback? Here's an example of getPastDiningBalances using async/await:
```
func getPastDiningBalances(diningToken: String, startDate: String) async -> Result<[DiningBalance], NetworkingError> {
        ...
        guard let (data, response) = try? await URLSession.shared.data(for: request), let httpResponse = response as? HTTPURLResponse, httpResponse.statusCode == 200 else {
            return .failure(.serverError)
        }
        let decoder = JSONDecoder()
        decoder.keyDecodingStrategy = .convertFromSnakeCase
        if let decodedBalances = try? decoder.decode(PastDiningBalances.self, from: data) {
            return .success(decodedBalances.balanceList)
        } else {
            return .failure(.parsingError)
        }
    }
```
Message me if you need any help!

aantenberg (changes requested):
Good work! One request is to use async/await for the getDiningPlanStartDate function. Definitely reach out to me if you have questions!

PennMobile/Dining/Networking + Cache/DiningAPI.swift
JHawk0224: Yea ill do that!

PennMobile/Dining/SwiftUI/DiningAnalyticsViewModel.swift
```diff
@@ -76,19 +77,41 @@ class DiningAnalyticsViewModel: ObservableObject {
...
+            switch planStartDateResult {
+            case .failure:
+                return
+            case .success(let planStartDate):
+                // If dining plan found, start prediction from the date dining plan started
+                startDollarBalance = (self.dollarHistory.first { $0.date == planStartDate }) ?? startDollarBalance
+                startSwipeBalance = (self.swipeHistory.first { $0.date == planStartDate }) ?? startSwipeBalance
+                // However, it's possible that people recharged dining dollars (swipes maybe?), and if so, predict from this date (most recent increase)
+                for (i, day) in self.dollarHistory.enumerated() {
+                    if i != 0 && day.date > planStartDate && day.balance > self.dollarHistory[i - 1].balance {
```
aantenberg, line 92: beautiful

aantenberg (approved):
Awesome, looks good to me!

PR #436, opened 2022-10-30
Fix notification ID updating API interaction

kunlizhang:
We were not interacting with the backend API correctly for updating the device token. These updates should now correctly update the device token after fetching the notification ID.


PennMobile/General/Networking + Analytics/UserDBManager.swift
```diff
@@ -46,7 +46,7 @@ class UserDBManager: NSObject, Requestable, SHA256Hashable {
             task.resume()
         }
     }
-
+
```
01jongmin, line 49: Does this spacing not cause a linter error?

PennMobile/General/Networking + Analytics/UserDBManager.swift
```diff
@@ -463,8 +463,44 @@ extension UserDBManager {
...
+                    completion(.failure(.parsingError))
+                }
+            }
+            task.resume()
+        }
+    }
+        
+    // Updates device token.
+    func savePushNotificationDeviceToken(deviceToken: String, notifId: Int, _ completion: (() -> Void)? = nil) {
+        let url = "https://pennmobile.org/api/user/notifications/tokens/\(String(notifId))"
```
01jongmin, line 503: You don't need to wrap notifId into a string, you should be able to simply do `\(notifId)` as a means to use the notifId in the URL string

PennMobile/Setup + Navigation/AppDelegate+NotificationExtension.swift
```diff
@@ -28,7 +28,13 @@ extension AppDelegate: UNUserNotificationCenterDelegate {
         }
 
         let token = tokenParts.joined()
-        UserDBManager.shared.savePushNotificationDeviceToken(deviceToken: token)
+
+        UserDBManager.shared.getNotificationId() { result in
```
01jongmin, line 32: I believe you wouldn't need the trailing `()`, in swift, if your last variable is a closure, you wouldn't need the closing brackets.

01jongmin (changes requested):
Couple of changes!

01jongmin (approved):
LGTM!

PR #437, opened 2022-10-31
Add Today's Courses widget

anli5005:
So, at long last, widgets.

Capture d’écran, le 2022-10-31 à 16 37 22

Implementing this required touching quite a few aspects of the code, which may or may not fundamentally change how the app is built:

* Some files have been moved to a Shared folder and are now shared across the app and its widget extension
  * The Shared folder contains an intent definition file, so in theory we could use it to implement Siri Shortcuts support (which would be pretty cool)
* A new `Storage.retrieveThrowing` method has been added to enable error handling
* An App Group (`group.org.pennlabs.PennMobile`) has been created
  * Use the `Storage` directories `groupDocuments` and `groupCaches` to write to the group container
  * Course data is now stored in the group container for the widget to use
    * Data is migrated on app launch - the widget will prompt users to open the app if the data is not migrated yet
  * We might want to consider having a different app group identifier for development/having a separate mechanism to distinguish files created by release and development versions of Penn Mobile
* **Provisioning files may have to be redistributed/retooled** since a bunch of entitlements and targets were added.
  * The simplest workaround for me has been to check the "Automatically manage signing" checkbox. This worked for me without any problems, though YMMV
    * Let me know if I should revert this
  * Our deployment tooling may have to be reworked to accommodate the new widget extension and its signing

A few other notes:

* Comments are relatively sparse -- hopefully the code should be mostly self-documenting, but I can add them if there's anything confusing here
* `CoursesProvider` and `IntentCoursesProvider` are implemented in a way that will allow for other widgets (like lock screen widgets or a weekly courses widget), hence all the protocols and shared code between them


01jongmin (approved):
LGTM!

PR #439, opened 2022-11-14
Added native dining menus

JHawk0224:
Note that currently, besides the bad design for the menu I made, there are some bugs with the menu. I believe these are mostly backend's problems, but I'll list them anyway.
• Some ingredients are duplicated
• Sometimes description is also duplicated
• I have seen a few instances where "<br />" is included in the ingredients list lmao
• On our end, when parsing the ingredients we split at ", " to get each ingredient, but there are some instances where this isn't correct, for example if the ingredients list is: "stock chicken house-made (water, chicken, carrots, celery, leeks, onion, garlic, bouquet garni), white rice parboiled no added salt (water, rice), andouille pork sausage, carrot, tomatoes, ..."
• I believe all the caching and getting menus from certain dates should work fine, but I wouldn't be surprised if there is some edge case where the displaying menu is from the wrong date
• The menu that is automatically showing (e.g. breakfast vs lunch vs dinner) I don't think always works if the meal is called brunch or something other than those three, but I haven't tested this yet

### DIFF
--- a/PennMobile/Dining/Model/DiningVenue+Extensions.swift
+++ b/PennMobile/Dining/Model/DiningVenue+Extensions.swift
@@ -70,6 +70,10 @@ extension DiningVenue {
     var currentOrNearestMealIndex: Int {
         return self.mealsToday?.meals.firstIndex(where: { $0.isCurrentlyServing }) ?? self.mealsToday?.meals.firstIndex(where: { $0.starttime > Date() }) ?? 0
     }
+    
+    var currentOrNearestMeal: Meal? {
+        return self.mealsToday?.meals.first(where: { $0.isCurrentlyServing }) ?? (self.mealsToday?.meals.first(where: { $0.starttime > Date() }) ?? nil)
+    }
 
     var hasMealsToday: Bool {
         return mealsToday != nil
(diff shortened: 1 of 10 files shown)


JHawk0224:
New menu design made by Joy

PennMobile/Dining/Model/DiningVenue+Extensions.swift
```diff
@@ -70,6 +70,10 @@ extension DiningVenue {
     var currentOrNearestMealIndex: Int {
         return self.mealsToday?.meals.firstIndex(where: { $0.isCurrentlyServing }) ?? self.mealsToday?.meals.firstIndex(where: { $0.starttime > Date() }) ?? 0
     }
+    
+    var currentOrNearestMeal: Meal? {
+        return self.mealsToday?.meals.first(where: { $0.isCurrentlyServing }) ?? (self.mealsToday?.meals.first(where: { $0.starttime > Date() }) ?? nil)
```
aantenberg, line 75: Nice!

Shared/General/Colors.swift
```diff
@@ -124,6 +124,7 @@ extension Color {
     static var grey4 = Color("grey4")
     static var grey5 = Color("grey5")
     static var grey6 = Color("grey6")
+    static var grey7 = Color("grey7")
```
aantenberg, line 127: Most important change

aantenberg (approved):
This looks good to me! I can tell you put a lot of work into this -- good job. If there are no outstanding questions about it I think it looks good to merge!

PR #442, opened 2022-12-04
Add localized descriptions for Siri intents

anli5005:
Should silence warnings from App Store Connect.


aantenberg (approved):
Hmm.. Might have to think about this one.
lgtm!!!

PR #444, opened 2022-12-09
Attempt to fix dining balance cache crash

anli5005:
My best guess is that this is being caused by some sort of race condition, where the widget and the app try to save data to the same cache file at the same time. This PR:

* Gracefully handles errors if that does happen by adding a new `storeThrowing` method
* Changes the `Storage.store` architecture to not rely on `fileExists`


01jongmin (approved):
Let's try and see if this is the fix!

PR #447, opened 2023-01-20
Dining analytics refresh fix

JHawk0224:


Shared/Dining/DiningAnalyticsViewModel.swift
```diff
@@ -50,7 +50,9 @@ class DiningAnalyticsViewModel: ObservableObject {
     init() {
         formatter.dateFormat = "yyyy-MM-dd"
         clearStorageIfNewSemester()
-        populateAxesAndPredictions()
+        Task {
+            await refresh(refreshWidgets: true)
+        }
```
anli5005, line 55: Maybe we should just delete this altogether? IMO it's probably a good idea to keep refreshing the caller's responsibility

#448 moves refreshing to outside the `.task` modifier, so we should be good on that front.

PR #451, opened 2023-02-05
Made new analytics graph with support for touching the graph

JHawk0224:
The functionality is complete, and it works well. Then only thing I think we need to change is how the display actually looks. We could make it display above/below the line depending on the quadrant, add a transparent background to the text, or anything else etc. We should probably get a designer to plan this, but I think it looks good for now.

Also, the new graphs only work in iOS 16, but only 75% of our users use iOS 16 currently. Because of this, I left all of the old code in and only display the new graph if iOS 16 is being used. Therefore all of the old code is still actually in use, and cannot be deleted yet. However, after more people switch over to iOS 16, we can delete every other file in ./PredictionsGraph/, and DiningAnalyticsView can be vastly simplified.

https://user-images.githubusercontent.com/35184414/216840137-06e541fe-e423-4330-8a7e-8c2868345840.mov


aantenberg:
Looks great!

aantenberg (approved):
Looks great! So happy with this change

Issue #452, opened 2023-02-16
Dining Dollars Prediction Skewed by Incorrect 0 Data Points


george-botros:
In the Dining Analytics, sometimes data points indicating that there are no more dining dollars are added, and this skews the line of best fit downward. I'm not sure, but, just from observation, these 0 data points seem to appear after the app makes you re-enter your PennKey. 
![IMG_3532](https://user-images.githubusercontent.com/78520093/219491349-1725493d-88ae-4091-bdde-a7cee20e78f9.PNG)

PR #454, opened 2023-02-17
Fixed bug with missing dining data on graph

JHawk0224:
Fixes issue #452

On Feb. 13 (and I think maybe 15th), some people have missing dining data. Some people literally don't have data for this day, while others have $0. This patch tries to remove outliers by removing all days where the balance is 0, but the day before and day after have positive balance. Note that this could be incorrect for people who refill their balances the same day they run out (which we assume is no one).

Also, now on the graph, if there is missing data, instead of nothing showing up, it displays the balance from the most recent day beforehand.


aantenberg (approved):
Looks good to merge!

PR #456, opened 2023-02-19
Add haptic feedback for graphs

anli5005:
Makes the graph far more satisfying to use.

Note that sometimes the haptic generator borks if you trigger haptics too quickly. Not sure why this happens, but it's not a major issue imo.


aantenberg (approved):
Looks awesome. Great idea!

PR #460, opened 2023-03-26
Polls!

kspadman:
Initial final commit for polls.

Issue with generating Authentication token in PollsNetworkManager. Used the same OAuth2NetworkManager.instance.getAccessToken as is used for other token generation in the app. In running on simulator, have had issues generating tokens. Don't expect this to be a problem on other devices because using same function to generate these tokens as other features on PennMobile!

Note: the "id_hash" field must be changed to be unique based on the user. This can be found in PollsNetworkManager, line 14. Currently just using a dummy string to get active polls. 

See below video for what it looks like when working (spacing is fixed in this PR)

https://user-images.githubusercontent.com/99234666/227801893-32fa8814-c661-4cd5-9dc7-d1bdc03af1dc.mov


JHawk0224 (approved):
We still need to add poll history, but for now it's okay to merge.

PR #461, opened 2023-03-29
Switch /token endpoint to pennmobile.org

anli5005:
This PR switches the token route to the Penn Mobile backend (as advised by @judtinzhang), which lets us do more interesting things on the Penn Mobile side.

Haven't thoroughly tested for regressions, but it looks like thing aren't entirely dead from cursory testing., The one wildcard is course alerts (which uses some sort of CSRF thing), but it's unclear whether that worked in the first place.


raunaqsingh2020 (approved):
💯 LGTM thanks for taking care of this boss @anli5005

PR #469, opened 2023-09-02
Make widgets compatible with the iOS 17 SDK

anli5005:
Applies `containerBackground` modifiers as necessary to make sure our widgets don't end up looking like this when we compile for iOS 17:

A widget saying "Please adopt containerBackground API"

Also tweaks widget paddings to use system-provided ones when possible.

As an aside, this enables StandBy support for all of our widgets!


JHawk0224 (approved):
LGTM

PR #471, opened 2023-09-03
Revert "Make widgets compatible with the iOS 17 SDK"

anli5005:
I guess Xcode 14 doesn't like us referencing methods it doesn't know about yet


JHawk0224 (approved):
.

PR #472, opened 2023-09-14
Revert "Revert "Make widgets compatible with the iOS 17 SDK""

anli5005:
Reverts pennlabs/penn-mobile-ios#471 now that iOS 17 is releasing within the next week.

If your build breaks due to this PR, upgrade to Xcode 15 or branch off an earlier commit.


JHawk0224 (approved):
Looks good

PR #473, opened 2023-09-15
Add laundry live activity

anli5005:
Adds live activities for tracking the progress of your laundry on the lock screen and Dynamic Island:

Laundry on the Dynamic Island

Two live activities on the lock screen

(Note that this PR is based on the changes in #472.)


JHawk0224 (approved):
Looks amazing! :D

PR #474, opened 2023-09-29
Manually handle signing

anli5005:
This PR disables automatic management of signing, in lieu of manually issuing provisioning profiles.

While this change has technically been made on everyone's computers, it's probably better to just get this in a PR rather than everyone committing it at the same time.


JHawk0224 (approved):
Looks good!

PR #479, opened 2023-10-04
Add Poll History page

JHawk0224:
This is ready to be merged. In the future, we should probably improve the ui, but I think we should go ahead and merge first, and fix the design later.

If you want to test it out, you can set the id in PollsNetworkManager to "E32FF1C8-BD57-4925-AB88-513208C9CA14" as that already has a vote that corresponds with it.

Simulator Screenshot - iPhone 15 Pro - 2023-10-04 at 12 50 55


PennMobile/Polls/PollsView.swift
```diff
@@ -0,0 +1,58 @@
...
+    func makeUIView(context: Context) -> HomePollsCell {
+        let cell = HomePollsCell()
+        cell.pollQuestion = answeredPollQuestion
+        cell.isUserInteractionEnabled = false
+        return cell
+    }
+    
+    @available(iOS 16.0, *)
+    func sizeThatFits(_ proposal: ProposedViewSize, uiView: HomePollsCell, context: Context) -> CGSize? {
+        return CGSize(width: proposal.width ?? 0, height: HomePollsCell.getPollHeight(for: answeredPollQuestion))
```
JHawk0224, line 52: Also I wanted to get rid of this "?? 0" and use the default size if possible, but I couldn't figure out how to. If nil is returned from this function, then the default size is used, but I couldn't figure out how to just set the height.

anli5005:
Can we convert this to a draft?

anli5005 (approved):
Will give it the stamp - ***however,*** check the comments for issues if we're trying to guarantee 100% anonymity between the client and the server.

PennMobile/Polls/Networking/PollsNetworkManager.swift
```diff
@@ -11,85 +11,99 @@ import SwiftyJSON
 import PennMobileShared
 
 class PollsNetworkManager: NSObject, Requestable {
-
     static let id = UIDevice.current.identifierForVendor?.uuidString ?? ""
```
anli5005, line 14: Looks like GitHub ate my comment earlier, but I see two issues with this approach to anonymizing response IDs:

1. In theory the backend could trace this back to a user's access token (since the device identifier is also sent with some other requests, such as laundry). In practice, the server doesn't store this data, but it might be better to just generate a random UUID when the app first launches
2. It looks like we're sending the access token with each poll request anyway in `answerPoll`

PR #482, opened 2023-10-08
made the slope average of total + last7

christinaqiu3:


JHawk0224 (changes requested):
Looks great! I'm a little worried in general though that some people won't like this change. I personally would like it if my predictions kept the slope from the whole semester. Do you think we should add an option somewhere to toggle between these?

On another note, I also personally want to test out the following slope calculation:
For every pair of data points, get the slope between them. Then find the weighted average of all these slopes (based on how many days apart they were from), and use this as the final slope.

PennMobileShared/Dining/DiningAnalyticsViewModel.swift
```diff
@@ -194,7 +209,10 @@ public class DiningAnalyticsViewModel: ObservableObject {
         return balanceDiff / timeDiff
     }
     func predictZeroDate(firstBalance: DiningAnalyticsBalance, lastBalance: DiningAnalyticsBalance, slope: Double) -> Date {
-        let offset = -firstBalance.balance / slope
+        var offset = -firstBalance.balance / slope
+        if slope == 0 {
```
joyliu-q, line 213: @JHawk0224 correct me if I'm wrong but I think you can also do:

return slope == 0 ? Date.distantFuture : Calendar.current.date(byAdding: .day, value: Int(offset), to: firstBalance.date)!

PennMobileShared/Dining/DiningAnalyticsViewModel.swift
```diff
@@ -108,43 +108,58 @@ public class DiningAnalyticsViewModel: ObservableObject {
 
     func populateAxesAndPredictions() {
         filterData()
+
```
joyliu-q, line 111: Some comment documentation about the different methods we use would be great! We're starting to use a variety of methods so clarity becomes important here for future development.

PennMobileShared/Dining/DiningAnalyticsViewModel.swift
joyliu-q: Not a swift guru lol but I think it still has the ternary operator syntax

JHawk0224 (approved):
Looks good, I think it's ready to merge in!

PR #485, opened 2023-10-11
fix: Fixed dining venues sometimes not appearing (hopefully)

george-botros:
As reported on [Airtable](https://airtable.com/appTHcLDue56TbPL4/tblJYBj92x5vmyHzW/viw12ToU39Gcym38M/recOox5W7NtiUt1lz?blocks=hide), sometimes dining venues do not load. This problem was attempted to be fixed by checking if the list of dining venues is empty, and, if so, refreshing. However, the implementation of this fix was incorrect. The dining venues list is guaranteed to be nonempty. If it is contains no _venues_, it will have two empty arrays. That is, diningVenues will equal `[PennMobileShared.VenueType.dining: [], PennMobileShared.VenueType.retail: []]` but diningVenues.isEmpty will be false. This PR fixes this implementation by checking if each category is empty.

### DIFF
--- a/PennMobile/Dining/SwiftUI/DiningViewModelSwiftUI.swift
+++ b/PennMobile/Dining/SwiftUI/DiningViewModelSwiftUI.swift
@@ -21,13 +21,20 @@ class DiningViewModelSwiftUI: ObservableObject {
     @Published var alertType: NetworkingError?
 
     @Published var diningBalance = (try? Storage.retrieveThrowing(DiningBalance.directory, from: .groupCaches, as: DiningBalance.self)) ?? DiningBalance(date: Date.dayOfMonthFormatter.string(from: Date()), diningDollars: "0.0", regularVisits: 0, guestVisits: 0, addOnVisits: 0)
+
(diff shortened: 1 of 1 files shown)


PennMobile/Dining/SwiftUI/DiningViewModelSwiftUI.swift
```diff
@@ -46,6 +46,12 @@ class DiningViewModelSwiftUI: ObservableObject {
         }
     }
 
+    func areAllVenuesEmpty(diningVenues: [VenueType: [DiningVenue]]) -> Bool {
```
anli5005, line 49: Why do we need to pass in `diningVenues` as a parameter? Could we just make this a computed property?

anli5005, line 49: Why do we need to pass in diningVenues as a parameter? Could we just make this a computed property?

JHawk0224 (commented):
Looks good

anli5005 (approved):

PR #486, opened 2023-10-15
Fix dynamic tabs drag bug

JHawk0224:


PennMobile/Preferences/PreferencesView.swift
```diff
@@ -35,6 +35,7 @@ struct PreferencesView: View {
                     Text("Choose which features appear in the tab bar. The remaining ones can still be found on the More tab.")
                 }
             }
+            .id(UUID())
```
anli5005, line 38: What bug was this fixing?

anli5005 (approved):

PennMobile/Preferences/PreferencesView.swift
JHawk0224: It fixes this weird visual glitch in the list when you click and hold for too long on one element
https://github.com/pennlabs/penn-mobile-ios/assets/35184414/e633e329-8501-4808-a155-da3f552530ee

PR #475, opened 2023-10-01
refactor: Moved shared code into PennSharedCode framework

george-botros:
Converted the shared code file into a framework called PennSharedCode. Modified functions/etc. to public as necessary within the framework. Made variable scoping changes to CoursesViewModel > extension Course.

PR #490, opened 2023-11-05
Fix course metadata not populating

anli5005:
Fixes regressions from #475 that could cause the course schedule view to display courses from all semesters at once and without locations


JHawk0224 (approved):
LGTM

PR #491, opened 2023-11-10
Add laundry ids to display

JHawk0224:
Screenshot 2023-11-09 at 9 36 52 PM

### DIFF
--- a/PennMobileShared/Laundry/LaundryMachine.swift
+++ b/PennMobileShared/Laundry/LaundryMachine.swift
@@ -67,20 +67,21 @@ public class LaundryMachine: Hashable, Codable {
 // MARK: - Comparable
 extension LaundryMachine: Comparable {
     public static func < (lhs: LaundryMachine, rhs: LaundryMachine) -> Bool {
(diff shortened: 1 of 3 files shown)


PennMobileShared/Laundry/LaundryMachine.swift
```diff
@@ -67,20 +67,21 @@ public class LaundryMachine: Hashable, Codable {
...
+//        case (_, .offline),
+//             (_, .outOfOrder):
+//            return true
+//        case (.offline, _),
+//             (.outOfOrder, _):
+//            return false
+//        default:
+//            return lhs.timeRemaining < rhs.timeRemaining
+//        }
+        return lhs.id < rhs.id
```
anli5005, line 84:
```suggestion
        lhs.id < rhs.id
```

We can always grab the previous code from the diff later if we need

anli5005 (approved):

PennMobileShared/Laundry/LaundryMachine.swift
JHawk0224: But then no one will know about it

PR #493, opened 2023-11-12
Aesthetic edits on anli/09-17-refactor

george-botros:
Removed less relevant links from more tab to reduce clutter.

Change capitalization of preferences in notifications and privacy tabs in more tab

PR #492, opened 2023-11-10
Refactor the core of the app using SwiftUI

anli5005:
The long-awaited refactor of our app in SwiftUI is here.

New home screen

Note that this PR updates the minimum deployment target to iOS 16.0.

Some tasks that I've deferred until later PRs:
* Rewriting the PAC code tab in SwiftUI (as that relies on navigation bar controls)
* ~~Fixing a rare crash when switching to the GSR/laundry tabs after re-pinning them to the tab bar (those are also related to navigation bar controls)~~ *now fixed!*
* Deleting now-unused code
* Refactoring the "Edit Features" screen to use the new `AppFeature` structs
* Tweaking the home screen with feedback from design

PR #494, opened 2023-11-12
Fix App Update Notifications

anli5005:
Fixes https://github.com/pennlabs/penn-mobile-ios/pull/493#discussion_r1390460968


anli5005:
I've lost track of which branch is which so we'll just make the changes on the original PR #492

PR #495, opened 2023-12-04
Upsert Path Registration in PCP upon fetching courses from Path

mureytasroc:
This PR adds an `updatePathRegistration` method to the existing `CourseAlertNetworkManager`, which maybe we should rename to something more generic to the Penn Courses backend (not important). This method can be used to upsert the user's Path registration as a PCP schedule. This method will be called as a side effect by `fetchStudentCourses`.

This PR also modifies the `getToken` method in `PathAtPennNetworkManager`. I don't think this method actually works currently. It pulls the user's pennkey and password from `KeychainAccessible`, and their Duo cookies from `PennLoginController`’s `WKWebsiteDataStore`. If the username/password are stored and the duo cookie isn't expired (check for cookie expiration before sending request so we don't trigger errant 2FA push notification), it initiates the Penn auth flow by sending a POST request to `https://weblogin.pennkey.upenn.edu/idp/profile/oidc/authorize?execution=e1s1&eventTag=password` with the following `application/x-www-form-urlencoded` content:
```
j_username=pennkey&j_password=password&_eventId_proceed=
```
(where `pennkey` is the user's pennkey and `password` is the user's password, both URLencoded)

This flow will eventually redirect to Duo, which uses a cookie with a key like `fsc|DU8WC5BNIIZ03YFXMNNK` to bypass 2FA (this cookie is set when the user clicks "remember this device" in the 2FA flow upon app login). If the auth flow fails for any reason, `getToken` raises and log the user out, so that the next time they login they will refresh their pennkey credentials and Duo cookie.

Finally, the auth flow returns the token in the same format currently assumed by the `getToken` function (so the parsing can stay the same).

Besides modifying `getToken` to actually work, this PR handles accepting a silent push notification from mobile backend(?) to trigger refreshing of a student's courses (say once per hour).

If we really think a user's Path course schedule is sensitive info (it's not imo), we should also add an opt-in popup so that the user's schedule isn't synced with PCP until/unless they grant permission. But imo since the data isn't rly sensitive enough to warrant off-server storage and won't be leaving the Penn Labs ecosystem anyway (which already stores sensitive info and is very careful about user privacy), this seems like unnecessary dev time and user burden for what it accomplishes. Def up for discussion.

### DIFF
--- a/PennMobile/Course Alerts/Networking/CourseAlertNetworkManager.swift
+++ b/PennMobile/Course Alerts/Networking/CourseAlertNetworkManager.swift
@@ -7,6 +7,7 @@
 //
 import Foundation
 import SwiftyJSON
+import PennMobileShared
 
 struct Response: Decodable {
     let message: String
@@ -159,26 +160,23 @@ class CourseAlertNetworkManager: NSObject, Requestable {
         }
     }
 
-    func updatePathRegistration(srcdb: String, crns: [String], callback: @escaping (_ success: Bool, _ error: Error?) -> Void) {
-        struct Section: Encodable {
-            var id: String // crn
-        }
+    func updatePathRegistration(srcdb: String, crns: [String]) async throws {
+        let params: [String: Any] = ["semester": srcdb, "sections": crns.map { ["id": $0] }]
 
-        let sections = crns.map { Section(id: $0) } 
-        let params: [String: Any] = ["semester": srcdb, "sections": sections]
-
-        makeAuthenticatedRequest(url: pathRegistrationURL, requestType: RequestType.PUT, params: params) { (data, status, error) in
-            guard let status = status as? HTTPURLResponse else {
-                callback(false, error)
-                return
+        return try await withCheckedThrowingContinuation { continuation in
+            makeAuthenticatedRequest(url: pathRegistrationURL, requestType: RequestType.PUT, params: params) { (data, response, error) in
+                if let error {
+                    continuation.resume(throwing: error)
+                    return
+                }
+                
+                guard let response = response as? HTTPURLResponse, response.statusCode == 200 else {
+                    continuation.resume(throwing: NetworkingError.serverError)
+                    return
+                }
+                
+                continuation.resume(returning: ())
             }
-
-            guard data != nil else {
-                callback(false, error)
-                return
-            }
-
-            callback(status.statusCode == 200, error)
         }
     }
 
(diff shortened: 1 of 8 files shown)


PennMobile/Login/PathAtPennNetworkManager.swift
```diff
@@ -45,6 +51,49 @@ extension PathAtPennNetworkManager {
...
+            let matches = str.getMatches(for: “value: \“(.*)\“”)
+            let token = try matches.first.unwrap(orThrow: PathAtPennError.noTokenFound)
+
+            return token
+        } catch {
+            // Logout
+            DispatchQueue.main.async {
+                AppDelegate.shared.rootViewController.clearAccountData()
+                AppDelegate.shared.rootViewController.switchToLogout()
+            }
```
anli5005, line 93: The behavior on failure should not be to log out the user, as this removes user data and can be disruptive. Can we just throw an error from here and let the caller present some sort of error message with an option to reauthenticate to Path@Penn specifically?

PennMobile/Course Alerts/Networking/CourseAlertNetworkManager.swift
```diff
@@ -158,6 +159,29 @@ class CourseAlertNetworkManager: NSObject, Requestable {
         }
     }
 
+    func updatePathRegistration(srcdb: String, crns: [String], callback: @escaping (_ success: Bool, _ error: Error?) -> Void) {
```
anli5005, line 162: Is it possible to make this an async method to make PathAtPennNetworkManager a bit cleaner? We can do this by moving the `withCheckedThrowingContinuation` inside here.

PennMobile/Login/PathAtPennNetworkManager.swift
```diff
@@ -149,6 +198,20 @@ extension PathAtPennNetworkManager {
...
+                    if let error = error {
+                        continuation.resume(throwing: error)
+                    } else if success {
+                        continuation.resume(returning: ())
+                    } else {
+                        // Handle the case where success is false and there's no error
+                        continuation.resume(throwing: NSError(domain: "UpdatePathRegistrationError", code: 0, userInfo: [NSLocalizedDescriptionKey: "Unknown error occurred"]))
+                    }
+                }
+            }
```
anli5005, line 213: Should we wrap this in a `Task`? This would make the function not wait for the registration to be updated before returning.

PennMobile/Login/PathAtPennNetworkManager.swift
```diff
@@ -45,6 +51,49 @@ extension PathAtPennNetworkManager {
 
         return token
     }
+
+    func getToken() async throws -> String {
```
anli5005, line 55: We should combine this with the previous method (I don't think this will even compile.) Maybe something like:

1. Try getting the token the old way
2. If that fails, try logging in with the saved username/password
3. If *that* fails, then we throw an error

anli5005:
PR should be good to merge now - I have removed the call to PCA (as we really should not be handling that inside PathAtPennNetworkManager), and I'll write up another PR that calls PCA based on the SwiftUI refactor.

JHawk0224 (approved):
Looks good, apologies for the delay in looking at this PR!

PR #497, opened 2023-12-15
Add support for toasts

anli5005:
Reimplements `addStatusBar` using SwiftUI:

https://github.com/pennlabs/penn-mobile-ios/assets/7005789/d722320c-68d6-49d7-ad6a-eb2e118a1a01

Note that UIKit integration is currently implemented in a hacky manner - the hope is that we'll no longer need it a few months down the line.


JHawk0224 (approved):
Thank you for doing this, looks great!

PR #499, opened 2023-12-17
Make project compile under Xcode 15.1

anli5005:
For some reason Xcode 15.1 began complaining about this. I've made this change already in #495 and #497, but figured I'd add it here so we can quickly get it merged in.


JHawk0224 (approved):
👍

PR #502, opened 2024-01-17
Add explanatory message for Dining Analytics not appearing before semester begins

anli5005:
Amends the "No Dining Plan" screen with the following text:

> Dining Analytics may not appear until the day after the semester begins.


JHawk0224 (approved):
Looks good, thank you 😃

PR #504, opened 2024-01-21
fix: indicate venue is closed when it has no hours in detailed hours views

george-botros:
Currently, when a venue is closed, it displays nothing under the week day. This PR makes it say "closed" instead.

Before:

After:


JHawk0224 (approved):
Thanks, looks good!

PR #506, opened 2024-01-28
Subletting dialogue completed "custom popup view"

christinaqiu3:


PennMobile/General/SwiftUI Views/CustomPopupView.swift
```diff
@@ -0,0 +1,109 @@
...
+//  PennMobile
+//
+//  Created by Christina Qiu on 1/26/24.
+//  Copyright © 2024 PennLabs. All rights reserved.
+//
+
+#if canImport(SwiftUI)
+import SwiftUI
+import PennMobileShared
+#endif
```
anli5005, line 12:
```suggestion
import SwiftUI
import PennMobileShared
```

The `canImport` check shouldn't be necessary, but we can deal with that in a fast follow

PennMobile/General/SwiftUI Views/CustomPopupView.swift
```diff
@@ -0,0 +1,109 @@
...
+struct CustomPopupView: View {
+    @Binding var isShown: Bool
+    var image: Image
+    var title: String
+    var message: String
+    var confirmAction: () -> Void
+    var body: some View {
+        ZStack {
+            // Scrim
+            Color(hex: 0x5F5F64)
```
anli5005, line 32: We should avoid hardcoding colors - either use an existing one or create one in the asset catalog.

PennMobile/General/SwiftUI Views/CustomPopupView.swift
```diff
@@ -0,0 +1,109 @@
...
+                Text(title).font(.headline)
+                Text(message)
+                    .font(.subheadline)
+                    .multilineTextAlignment(.leading)
+                Button(action: confirmAction) {
+                    Text("Confirm")
+                        .foregroundColor(.white)
+                        .padding()
+                        .frame(maxWidth: 300)
+                        .background(RoundedRectangle(cornerRadius: 50).fill(Color.blue))
```
anli5005, line 52: Use `Color("navigation")` or define a new custom color in the asset catalog instead? This doesn't quite match the blue in the Figma

Screenshot 2024-02-03 at 17 10 36

anli5005: Same for the icon above

PennMobile/General/SwiftUI Views/CustomPopupView.swift
```diff
@@ -0,0 +1,109 @@
...
+                }
+                .padding(.horizontal, 24)
+                Button("Cancel") {
+                    isShown = false
+                }
+            }
+            .frame(maxWidth: 280)
+            .padding(.horizontal, 24)
+            .padding(.vertical, 32)
+            .background(Color.white)
```
anli5005, line 62: `Color("uiCardBackground")` should work for both Light Mode and Dark Mode

PennMobile/General/SwiftUI Views/CustomPopupView.swift
```diff
@@ -0,0 +1,109 @@
...
+extension Color {
+    init(hex: Int, alpha: Double = 1.0) {
+        self.init(
+            red: Double((hex >> 16) & 0xFF) / 255.0,
+            green: Double((hex >> 8) & 0xFF) / 255.0,
+            blue: Double(hex & 0xFF) / 255.0,
+            opacity: alpha
+        )
+    }
+}
```
anli5005, line 100: Should we remove this for now? See above comment about not hardcoding colors

(This'll still be in the commit history so we can revive it should, say, we need to use dynamic colors from the backend.)

PennMobile/General/SwiftUI Views/CustomPopupView.swift
```diff
@@ -0,0 +1,109 @@
...
+                CustomPopupView(isShown: $popupManager.isShown, 
+                                image: popupManager.image,
+                                title: popupManager.title,
+                                message: popupManager.message,
+                                confirmAction: {
+                                    // Define the confirm action here
+                                    popupManager.isShown = false
+                                })
+                .transition(.scale)
+            }
```
anli5005, line 86: Can we move this into `RootView`? Having it in a separate `ContentView` is arguably unnecessary, and it lines up with what we do with toasts

anli5005: You can also use `.overlay` since that implicitly uses a `ZStack` in case that makes the code cleaner

PennMobile/General/SwiftUI Views/DialogueViewController.swift
```diff
@@ -0,0 +1,66 @@
...
+    func hide() {
+        UIView.animate(withDuration: 1, delay: 0.0, options: .curveEaseOut) {
+            self.backView.alpha = 0
+            self.contentView.alpha = 0
+        } completion: { _ in
+            self.dismiss(animated: false)
+            self.removeFromParent()
+        }
+    }
+}
```
anli5005, line 66: Let's remove this file (and the `.xib`) for now since they're UIKit - we can re-add them if the need arises

PennMobile/Setup + Navigation/PennMobile.swift
```diff
@@ -14,7 +14,8 @@ struct PennMobile: App {
     @UIApplicationDelegateAdaptor var delegate: AppDelegate
     @ObservedObject var authManager = AuthManager()
     @ObservedObject var homeViewModel = StandardHomeViewModel()
-    
+    @StateObject var popupManager = PopupManager()
```
anli5005, line 17: `PopupManager` should be declared on `RootView` and not the `App` struct itself - declaring them here will mean every window of the app (on iPadOS or macOS) shares the same popup state.

PennMobile/Setup + Navigation/PennMobile.swift
```diff
@@ -53,6 +54,9 @@ struct PennMobile: App {
                 .environmentObject(mockHomeViewModel)
             #endif
                 .accentColor(Color("navigation"))
+            
+            ContentView()
+                .environmentObject(popupManager)
```
anli5005, line 59: Let's move this into `RootView` - we can add an `.overlay` modifier like we do for toasts to keep the code cleaner here.

PennMobile/General/SwiftUI Views/CustomPopupView.swift
```diff
@@ -0,0 +1,109 @@
...
+                    .resizable()
+                    .aspectRatio(contentMode: .fit)
+                    .frame(width: 50, height: 50)
+                    .foregroundColor(.blue)
+                Text(title).font(.headline)
+                Text(message)
+                    .font(.subheadline)
+                    .multilineTextAlignment(.leading)
+                Button(action: confirmAction) {
+                    Text("Confirm")
```
anli5005, line 48: Users of `CustomPopupView` should be able to control the number and text of these buttons - in the design, [one dialog has only one "See My Listings" button](https://www.figma.com/file/Fy0QlnbJFpbxQIkRWTtwJY/Subletting?type=design&node-id=642-13224&mode=design&t=tg8AXIu7r2nzeoxk-4), which we should support:

Screenshot 2024-02-03 at 17 30 32

anli5005 (commented):
Looking good! Like the design so far, but there's some implementation details that should be worked out:

anli5005 (approved):
Awesome, looks great!

PR #508, opened 2024-02-17
Mass deletion of old files no longer needed

JHawk0224:
Marking as a draft, but making so we remember to merge this in before pushing the next version


JHawk0224:
@anli5005 Can you look this over and see if it's okay to merge in?

PR #512, opened 2024-03-15
Improve URL and image handling for NewListingForm

anli5005:
This PR:
* Compresses draft images so that saving drafts don't freeze the app for 5 seconds (temporary solution until drafts can be saved on backend)
* Add URL format validation and make external link field optional


JHawk0224 (approved):
Thank you. Looks great!

PR #513, opened 2024-03-16
Rearchitect HomeViewData

anli5005:
THis PR rearchitects `HomeViewModel` and `HomeViewData` so that content appears as it loads. This improves how responsive the app appears to be, even if it does lead to some layout shift when the app first loads.


JHawk0224:
LGTM

JHawk0224 (approved):
LGTM

PR #514, opened 2024-03-17
Fix navigation warning (the yellow warning screen) when sending interest

JHawk0224:
I haven't been able to test this because I can't log into Penn Mobile on my simulator (a separate issue), so can someone else test this for me? Just indicate interest in a sublet, and then clicking the "See Applied" should take you to the applied screen instead of showing a warning. Test this both when subletting is on the more tab and when it's in the nav bar.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -276,6 +276,7 @@ GEM
 PLATFORMS
   universal-darwin-21
   universal-darwin-22
+  universal-darwin-23
   x86_64-darwin-19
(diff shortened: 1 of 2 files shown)


anli5005 (approved):
LGTM! Seems to work from the More page.

Wasn't able to test while it was in the tab bar/saved screen as there's only one sublet not created by me so far, but it looks like it shouldn't break anything at the very least so I'm good with merging

PR #515, opened 2024-03-20
Tweak external link and form component behavior

anli5005:
This PR:
* Uses `SFSafariViewController` to present a sublet's external links
* Prevents malformed external links from crashing the app
* Makes form components not be red on first open for new sublets


JHawk0224 (approved):
This looks great! Do you think we also want to fix the validation red parts on the InterestForm view?

PR #518, opened 2024-03-22
Remove duplicate WebView

anli5005:
Quick PR to get main building again


anli5005:
For reference the duplicate view in question is https://github.com/pennlabs/penn-mobile-ios/blob/5b536bd7030e770c3203ae272f152d309b514224/PennMobile/Dining/SwiftUI/Views/Venue/Detail%20View/DiningVenueDetailView.swift#L150

JHawk0224 (approved):
LGTM

PR #519, opened 2024-03-23
Tweak dining menu header design

anli5005:
Adjusts padding and applies a bar material background to the dining menu header so it feels a bit more at home:

https://github.com/pennlabs/penn-mobile-ios/assets/7005789/192eee4b-dd58-45a0-a870-bacd0ca10907


jonathanmelitski (approved):
LGTM

Probably still need to do timeline logic but that can came later.

JHawk0224 (approved):
Looks good

PR #520, opened 2024-03-24
Fix crash from unwrapping optional in GSRController

anli5005:
We were getting a few crashes due to this.

Crash log from unwrapping sorted.first!


JHawk0224 (approved):
Love it

anli5005:
A lot of crashes, actually
454 devices over the last year

PR #521, opened 2024-03-24
Fix Listing Name field weirdness

anli5005:
Applies the following changes:
* Ensures that sublet view model updates always occur on the main thread, which should eliminate sources of deadlock/race conditions/bugs in general
* Updates PennForms to the `subletting` branch (which is intended to become the de facto main branch for subletting purposes) in order to fix not being able to tap on the Listing Name field
* Applies a ~~hack~~ carefully studied approach to ensure that the Listing Name field always populates ~~sh ignore the asyncAfter we totally don't just give up and load everything 1 ms after the view appears~~

### DIFF
--- a/PennMobile/Subletting/Listings/NewListingForm.swift
+++ b/PennMobile/Subletting/Listings/NewListingForm.swift
@@ -67,7 +67,7 @@ struct NewListingForm: View {
     init(sublet: Sublet) {
         self.isNew = false
         self.originalSublet = sublet
-        self._subletData = State(initialValue: sublet.data)
+        self._subletData = State(initialValue: SubletData())
         self._negotiable = State(initialValue: sublet.negotiable)
         self._price = State(initialValue: sublet.price)
         self._startDate = State(initialValue: sublet.startDate.date)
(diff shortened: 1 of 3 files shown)


PennMobile/Subletting/Listings/NewListingForm.swift
```diff
@@ -358,6 +358,18 @@ struct NewListingForm: View {
             }
         }
         .environment(\.showValidationErrors, showValidationErrors)
+        .onAppear {
+            // HACK: Listing Name field shows really weird behavior unless we do this??
+            if !isNew {
+                DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(1)) {
+                    if let originalSublet {
```
JHawk0224, line 365: Idk what to say about this

JHawk0224 (approved):
Ready to merge I think..?..

PennMobile/Subletting/Listings/NewListingForm.swift
anli5005: There is nothing to say about this.

This code is perfect.

It is the pinnacle of iOS development.

In the future, long after we have departed, this will be our legacy.

PR #522, opened 2024-03-24
Tweak home page error screen

anli5005:
Clears errors during refreshes and tweaks home view errors to look like this:

Error page with blue screen of death icon


innocentoak (approved):
LGTM!!

PR #524, opened 2024-03-31
Added Penn Mobile AI, the greatest innovation in the history of Penn Labs

jonathanmelitski:
Added Penn Mobile AI:

- It has a list of set words that bypass the text moderation
- It uses text moderation
- It's kinda rude ngl


JHawk0224 (approved):
This is amazing

PR #533, opened 2024-04-14
Prevent activityIndicatorBackground force unwrapping crash

anli5005:
Prevents a crash in case `viewDidLayoutSubviews()` is called before `configureAndLoad()`.


JHawk0224 (approved):
LGTM

PR #535, opened 2024-04-14
Fix getMenu() crash and revamp logic

anli5005:
* Ensures that `getMenu()` won't crash due to an index out of bounds error
* Attempts to match menus by name before index

### DIFF
--- a/PennMobile/Dining/SwiftUI/Views/Venue/Detail View/DiningVenueDetailMenuView.swift
+++ b/PennMobile/Dining/SwiftUI/Views/Venue/Detail View/DiningVenueDetailMenuView.swift
@@ -46,13 +46,29 @@ struct DiningVenueDetailMenuView: View {
     /// If the selected date is not the current day, return the first menu.
     /// If at any point, the list of menus is empty, return nil.
     func getMenu() -> DiningMenu? {
-        if (menus.count == 0) { return nil }
+        if menus.isEmpty {
+            return nil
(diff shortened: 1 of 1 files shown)


PennMobile/Dining/SwiftUI/Views/Venue/Detail View/DiningVenueDetailMenuView.swift
```diff
@@ -46,13 +46,29 @@ struct DiningVenueDetailMenuView: View {
     /// If the selected date is not the current day, return the first menu.
     /// If at any point, the list of menus is empty, return nil.
     func getMenu() -> DiningMenu? {
-        if (menus.count == 0) { return nil }
+        if menus.isEmpty {
+            return nil
+        }
         
-        if (!Calendar.current.isDate(menuDate, inSameDayAs: Date())) {
```
jonathanmelitski, line 51: bro leave me alone i come from java land

jonathanmelitski (approved):
LGTM, thanks for doing that. I really couldn't think of a way without just looking at specific cases where it crashes.

PennMobile/Dining/SwiftUI/Views/Venue/Detail View/DiningVenueDetailMenuView.swift
anli5005: no

PR #537, opened 2024-04-18
:memo: Update Airtable Link

joyliu-q:
Use 2023-2024 airtable link, not outdated one

### DIFF
--- a/PennMobile/More Tab/MoreView.swift
+++ b/PennMobile/More Tab/MoreView.swift
@@ -28,7 +28,7 @@ private struct PennLink: View, Identifiable {
 
 private let pennLinks: [PennLink] = [
     PennLink(title: "Penn Labs", url: "https://pennlabs.org"),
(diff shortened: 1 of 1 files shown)


PennMobile/More Tab/MoreView.swift
```diff
@@ -28,7 +28,7 @@ private struct PennLink: View, Identifiable {
 
 private let pennLinks: [PennLink] = [
     PennLink(title: "Penn Labs", url: "https://pennlabs.org"),
-    PennLink(title: "Share Your Feedback", url: "https://airtable.com/shrS98E3rj5Nw1wy6")]
+    PennLink(title: "Share Your Feedback", url: "https://airtable.com/appFRa4NQvNMEbWsA/shrVrNXT4PpI7nqvD")]
```
anli5005, line 31: Can we use https://pennlabs.org/feedback/ios instead so it's easier to update? We did this in #496 but it would seem I screwed up the links during the refactor.

```suggestion
    PennLink(title: "Share Your Feedback", url: "https://pennlabs.org/feedback/ios")]
```

anli5005 (changes requested):

anli5005 (approved):
LGTM

PR #545, opened 2024-04-22
Remove Spring Fling to reduce clutter

anli5005:
All this code is commented out, so we may as well delete it

We can always restore it from the Git history should we need (or just rewrite it in SwiftUI)


JHawk0224 (approved):
Do we know what this was for and if whatever it was we want to revive it?

PR #550, opened 2024-08-30
feat: dismiss sublet marketplace search keyboard immediately when scrolling

george-botros:
This PR dismisses the keyboard, which appears when you search for a sublet, immediately upon scrolling.

https://github.com/user-attachments/assets/56397a1f-2366-44f5-90f1-346580728cec

### DIFF
--- a/PennMobile/Subletting/MarketplaceView.swift
+++ b/PennMobile/Subletting/MarketplaceView.swift
@@ -203,7 +203,7 @@ struct MarketplaceView: View {
                 await sublettingViewModel.populateSublets()
                 await sublettingViewModel.populateFiltered()
             }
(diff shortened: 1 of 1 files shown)


PennMobile/Subletting/MarketplaceView.swift
```diff
@@ -203,6 +203,7 @@ struct MarketplaceView: View {
                 await sublettingViewModel.populateSublets()
                 await sublettingViewModel.populateFiltered()
             }
+            .scrollDismissesKeyboard(.immediately)
```
JHawk0224, line 206: Can we do interactively instead? I personally like it more

JHawk0224 (approved):
Looks good!

PR #552, opened 2024-09-01
feat: add circle indicators for swiping images

george-botros:
This PR introduces indicators for when images in a sublet detail view are swipeable. 

Max Images (6):

Swiping:

<= 1 Images:


JHawk0224 (approved):
LGTM!

PR #553, opened 2024-09-01
Add more options for candidate contact info

george-botros:
This PR adds a copy option to the email button in the sublet candidates view as well as a text and copy button to the phone number button. It also adds an email icon to the left of the email.

# Before: 
Before

# After:
After 1
After 2
After 3


JHawk0224 (approved):
Looks great, thanks for adding!

PR #556, opened 2024-10-11
Convert folder references back to groups

anli5005:
Hopefully should get CI working again because CocoaPods *still* doesn't support Xcode 16's synchronized folder references: CocoaPods/CocoaPods#12456


JHawk0224 (approved):
LGTM

PR #561, opened 2024-10-11
Fix dining menu index calculation logic

jonathanmelitski:
Fixed: 
- Calculation of current dining menu index
- Change date values in extensions to better reflect time zones

### DIFF
--- a/PennMobile/Dining/SwiftUI/Views/Venue/Detail View/DiningVenueDetailMenuView.swift
+++ b/PennMobile/Dining/SwiftUI/Views/Venue/Detail View/DiningVenueDetailMenuView.swift
@@ -28,7 +28,7 @@ struct DiningVenueDetailMenuView: View {
     
     @Binding private var parentScrollOffset: CGPoint
     
-    init(menus: [DiningMenu], id: Int, venue: DiningVenue, menuDate: Date = Date(), parentScrollProxy: ScrollViewProxy, parentScrollOffset: Binding<CGPoint>) {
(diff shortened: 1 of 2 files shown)


PennMobile/Dining/SwiftUI/Views/Venue/Detail View/DiningVenueDetailMenuView.swift
```diff
@@ -100,19 +101,18 @@ struct DiningVenueDetailMenuView: View {
...
-                menuDate = newDate
+                await diningVM.refreshMenus(cache: true, at: menuDate)
                 menus = diningVM.diningMenus[venue.id]?.menus ?? []
                 currentMenu = getMenu()
             }
         }
         
+        .onChange(of: currentMenu) { _ in
+            print((currentMenu?.service ?? "no menu") + " on " + menuDate.description)
+            selectedStation = nil
```
JHawk0224, line 114: Just wondering why we're setting this to nil?

JHawk0224 (approved):
Looks good (other than the comment below)

PennMobile/Dining/SwiftUI/Views/Venue/Detail View/DiningVenueDetailMenuView.swift
jonathanmelitski: SelectedStation is connected to the scrollviewproxy, so changing it causes an immediate scroll, whereas setting it to nil allows this value to be automatically adjusted by the user scrolling.

https://github.com/user-attachments/assets/575e0870-c2c7-4590-b5a8-f2cd1d1671a5

PR #567, opened 2024-11-10
Rewrite OAuth2NetworkManager as an actor

anli5005:
Fully ports `OAuth2NetworkManager` to async/await as a global actor. Also cleans up the logic to hopefully prevent edge cases where users were getting signed out due to overlapping refresh token requests.


JHawk0224 (approved):
Looks good!

PR #568, opened 2024-11-10
Convert PennMobileShared to Swift 6

anli5005:
Introduces strict concurrency checking to Penn Mobile's shared library.

I tried to do this with the widget extension as well, but it doesn't seem WidgetKit plays terribly well with concurrency.


JHawk0224 (approved):
Looks good!

PR #571, opened 2024-11-18
Fix Crashlytics script behavior when spaces are in directory names

anli5005:
Ran into this when I tried moving my derived data folder to an external drive.


JHawk0224 (approved):
LGTM

PR #573, opened 2024-11-19
Add time sensitive notification entitlement

anli5005:
Bumps the version number and adds the time sensitive notification entitlement for urgent alerts such as PCA.


JHawk0224 (approved):
Looks good! Is there a reason we need to bump the version number again though?

anli5005:
Yeah, it's because 8.0.7 is already in review so want to get a new TestFlight build out ASAP

Issue #581, opened 2024-12-01
Penn Mobile MacOS Sign-in Window is Misaligned


albert239825:
Not sure if you guys support MacOS, but I often used the MacOS version of PennMobile when my phone is not around. However, for signing into campus expresses for dining analytics, the sign-in window is not centered no matter how I resize the window. Thanks!

![image](https://github.com/user-attachments/assets/685b0166-51e1-4388-9803-512f45b7c803)

PR #584, opened 2025-02-06
Fix Campus Express login Mac bug

JHawk0224:
Addresses issue https://github.com/pennlabs/penn-mobile-ios/issues/581

Note there is still a bug on both versions of this page where canceling and reopening the sheet then stops the loading bar from showing (and then stale web request is still possible).


jonathanmelitski (approved):
LGTM, merge if you've tested locally.

PR #585, opened 2025-02-15
Fix authentication URL

jonathanmelitski:
Investigated a URL authentication flow since our current version was not functional.

Found the correct URL and replaced it. The rest of the authentication flow did not require any modification.

### DIFF
--- a/PennMobile/Login/PathAtPennNetworkManager.swift
+++ b/PennMobile/Login/PathAtPennNetworkManager.swift
@@ -45,11 +45,16 @@ class PathAtPennNetworkManager {
 // MARK: - Path@Penn Authentication
 
 extension PathAtPennNetworkManager {
-    static private let oauthURL = URL(string: "https://idp.pennkey.upenn.edu/idp/profile/oidc/authorize?response_type=code&scope=openid+email+profile&client_id=courses.upenn.edu%2Fsam_rMReby8Vl3M1BiyVWMAT&redirect_uri=https%3A%2F%2Fcourses.upenn.edu%2Fsam%2Fcodetotoken")!
+    static private let oauthURL = URL(string: "https://3scale-public-prod-path-authentication-anon.apps.k8s.upenn.edu/api/v1/leepfrogAuthn/login")!
 
     private func getTokenWithoutReauthenticating() async throws -> String {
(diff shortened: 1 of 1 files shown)


PennMobile/Login/PathAtPennNetworkManager.swift
```diff
@@ -45,11 +45,16 @@ class PathAtPennNetworkManager {
...
 
 extension PathAtPennNetworkManager {
-    static private let oauthURL = URL(string: "https://idp.pennkey.upenn.edu/idp/profile/oidc/authorize?response_type=code&scope=openid+email+profile&client_id=courses.upenn.edu%2Fsam_rMReby8Vl3M1BiyVWMAT&redirect_uri=https%3A%2F%2Fcourses.upenn.edu%2Fsam%2Fcodetotoken")!
+    static private let oauthURL = URL(string: "https://3scale-public-prod-path-authentication-anon.apps.k8s.upenn.edu/api/v1/leepfrogAuthn/login")!
 
     private func getTokenWithoutReauthenticating() async throws -> String {
         let (data, _) = try await URLSession.shared.data(from: PathAtPennNetworkManager.oauthURL)
         let str = try String(data: data, encoding: .utf8).unwrap(orThrow: PathAtPennError.corruptString)
+        let body = try SwiftSoup.parse(str)
+        let tokenContaining = try body.select("script").first().unwrap(orThrow: PathAtPennError.noTokenFound(str))
```
anli5005, line 54: Where is this used?

anli5005 (approved):
Going to give it a stamp but don’t understand some of the logic here

PennMobile/Login/PathAtPennNetworkManager.swift
jonathanmelitski: @anli5005 Took a while to figure out. The way they do authtokens now is a bit different (it opens a window that calls a function in its parent when it is done authenticating). That function is in a `script` tag in the HTML.

Take a look at the new URL (open in your browser), then look at the raw HTML that it returned after the auth flow completes. The working token (the one that we're after) is the constant stored in the script.

jonathanmelitski: image

PR #595, opened 2025-03-08
iOS 17

jonathanmelitski:
Based on existing analytics, the Penn Mobile user base does not have many iOS 16 users remaining. Further, many changes to Swift were introduced in iOS 17 to which we don't have access.

**This Pull Request:**

- Bumps minimum iOS Deployment Version to iOS 17 on all modules/targets.
- Bumps version number for CI/CD
- Fixes some deprecation warnings as a result of the new iOS Version (specifically `.onChange()` listeners, which went from one argument to zero or two (old + new) with this iOS version).

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
-The MIT License (MIT)
+MIT License
 
-Copyright (c) 2018 Penn Labs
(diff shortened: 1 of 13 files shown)


JHawk0224 (approved):
LGTM!

PR #596, opened 2025-03-11
Make toasts dismissable by swiping them upwards

jonathanmelitski:
Toast pop-ups should be dismissable. Given they pop downward from the top of the screen, the natural gesture to dismiss them is (I've seen people do this) swiping them upwards and away.

https://github.com/user-attachments/assets/24336ce8-2086-41ef-8da9-bffee7c97c57


JHawk0224 (approved):
Looks great!

PR #597, opened 2025-04-02
Update PennForms version to follow main branch

jonathanmelitski:
PennForms uses an external library called swiftui-introspect, which "requires" definition of the exact version to introspect (to ensure SwiftUI doesn't break things spontaneously). However, this requirement can be overridden, which has been done in PennForms.

https://github.com/pennlabs/PennForms/pull/7


JHawk0224 (approved):
Looks good

PR #602, opened 2025-04-30
PennWrapped

jonathanmelitski:


PennMobile/Home/HomeViewData.swift
```diff
@@ -16,12 +16,15 @@ struct HomeViewData {
...
     var posts: Result<[Post], Error>?
     var newsArticles: Result<[NewsArticle], Error>?
+    var wrapped: Result<WrappedModel, Error>?
     var events: [CalendarEvent] = []
     
     var onPollResponse: ((Int, Int) -> Void)?
     var onStartSubletting: (() -> Void)?
     var onDismissSubletting: (() -> Void)?
     
+    static let wrappedSemesters = "wrappedSemester."
```
jonathanmelitski, line 26: Remnant of WrappedSemesterState

PennMobile/Home/HomeViewData.swift
```diff
@@ -70,6 +73,16 @@ struct HomeViewData {
                     .padding(.bottom)
             }
             
+            if case .some(.success(let wrappedData)) = self.wrapped {
+                // TODO WRAPPED
+                // this pages.count could later become if wrappedData.pages != nil (pages being optional)
+                if wrappedData.pages.count > 0 {
+                    WrappedHomeScreenExperience(with: wrappedData)
+                }
```
jonathanmelitski, line 81: Need to implement ID field

PennMobile/Home/HomeViewData.swift
```diff
@@ -70,6 +73,16 @@ struct HomeViewData {
                     .padding(.bottom)
             }
             
+            if case .some(.success(let wrappedData)) = self.wrapped {
+                // TODO WRAPPED
```
jonathanmelitski, line 77: anything else to todo?

PennMobile/Setup + Navigation/RootView.swift
```diff
@@ -14,6 +14,7 @@ struct RootView: View {
     @State var toast: ToastConfiguration?
     @StateObject var popupManager = PopupManager()
     @Environment(\.scenePhase) var scenePhase
+    @State var wrappedActive: Bool = false
```
jonathanmelitski, line 17: get rid of this

PennMobile/Wrapped/WrappedExperience.swift
```diff
@@ -0,0 +1,33 @@
...
+    }
+    
+    public var body: some View {
+        if let active = stages.first(where: {$0.activeState == vm.wrappedExperienceState}) {
+            Group {
+                if let cVM = vm.containerVM {
+                    AnyView(active)
+                        .environmentObject(cVM)
+                } else {
+                    AnyView(active)
```
jonathanmelitski, line 26: this is a source of error. should be EmptyView instead

PennMobile/Wrapped/WrappedUnitView.swift
```diff
@@ -0,0 +1,47 @@
...
+    
+    @EnvironmentObject var experienceVM: WrappedExperienceViewModel
+    @ObservedObject var vm: WrappedContainerViewModel
+    
+    var body: some View {
+        GeometryReader { proxy in
+            LottieView(animation: unit.lottie)
+                .textProvider(DictionaryTextProvider(unit.values))
+                .playbackMode(vm.activeUnit == unit ? vm.activeUnitPlaybackMode : .paused(at:.currentFrame))
+                .currentProgress(vm.activeUnit == unit ? (CGFloat(vm.activeUnitProgress) * (unit.time / unit.lottie.duration) - floor(CGFloat(vm.activeUnitProgress) * (unit.time / unit.lottie.duration))) : 0)
```
jonathanmelitski, line 22: this _works_, but I need to recognize that this makes no sense.

PennMobile/Wrapped/WrappedUnitView.swift
```diff
@@ -0,0 +1,47 @@
...
+                .rotation3DEffect(
```
jonathanmelitski, line 23: this rotation effect is left as a remnant of when swipe actions were allowed, but is no longer used but could still be left in. Can be re-implemented using .allowsHitTesting for wrappedUnitView in the Container TabView.

PennMobile/Wrapped/WrappedUnitView.swift
jonathanmelitski: Like I don't know what this does

jonathanmelitski:
@JHawk0224 requested review, but there are some preemptive changes that anticipate the 2025S backend implementation of wrapped, so I don't yet have confirmation that this works. However, much of the review can be conducted regardless.

PennMobile/Home/HomeViewModel.swift
```diff
@@ -162,6 +162,41 @@ extension Optional {
...
+                        self.data.wrapped = .success(WrappedModel(semester: "", pages: []))
+                    }
+                    continuation.resume()
+                    return
+                }
+
+                let url = URLRequest(url: URL(string: "https://pennmobile.org/api/wrapped/semester/2025S/")!, accessToken: token)
+
+                let task = URLSession.shared.dataTask(with: url) { data, response, _ in
+                    guard let httpResponse = response as? HTTPURLResponse, let data = data, httpResponse.statusCode == 200 else {
```
JHawk0224, line 178: small thing that doesn't need to be blocked/changed, but you can just say `..., let data, ...` (as of [Swift 5.7](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0345-if-let-shorthand.md))

PennMobile/Wrapped/WrappedModel.swift
```diff
@@ -0,0 +1,61 @@
...
+public struct WrappedModel: Codable {
+    let semester: String
+    // Designed to be optional for forwards compatability
+    // (making pages an optional field was a design discussion for disabling wrapped between semesters)
+    let pages: [WrappedAPIUnit]?
+    
+    public init(from decoder: Decoder) throws {
+        let values = try decoder.container(keyedBy: CodingKeys.self)
+        self.semester = try values.decode(String.self, forKey: .semester)
+        let pagesRaw = try values.decodeIfPresent(SafeWrappedArray.self, forKey: .pages)
```
JHawk0224, line 20: Does the default decoder initialization not handle this? It looks like it would be the same logic but maybe I'm missing something.

PennMobile/Wrapped/WrappedAPIUnit.swift
```diff
@@ -0,0 +1,52 @@
...
+    let time: TimeInterval?
+    let lottieUrl: URL
+    let values: [String:String]
+    let name: String
+    
+    public init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        let durationString = try container.decodeIfPresent(String.self, forKey: .time)
+        self.id = try container.decode(Int.self, forKey: .id)
+        self.time = durationString != nil ? WrappedAPIUnit.timeInterval(from: durationString!) : nil
```
JHawk0224, line 28: Cool syntactical thing you can do, but honestly might be worse for code readability:
```swift
self.time = durationString.map { WrappedAPIUnit.timeInterval(from: $0) }
```
(Optionals have a `.map` function)

PennMobile/Wrapped/WrappedAPIUnit.swift
```diff
@@ -0,0 +1,52 @@
...
+
+public struct WrappedAPIUnit: Identifiable, Codable {
+    public static func == (lhs: WrappedAPIUnit, rhs: WrappedAPIUnit) -> Bool {
+        lhs.id == rhs.id
+    }
+    
+    public let id: Int
+    let time: TimeInterval?
+    let lottieUrl: URL
+    let values: [String:String]
```
JHawk0224, line 21: I'm assuming these are the key-value data used in Wrapped?

PennMobile/Wrapped/WrappedModel.swift
```diff
@@ -0,0 +1,61 @@
...
+        
+
+        while !container.isAtEnd {
+            do {
+                let element = try container.decode(WrappedAPIUnit.self)
+                tempElements.append(element)
+            } catch {
+                // The decoder only advances elements on a successful call, so calling decodeNil advances the
+                // decoder in the event that one is unsuccessful.
+                _ = try? container.decode(Empty.self)
```
JHawk0224, line 56: So why does the default decoder initialization not work again?

PennMobile/Wrapped/WrappedUnit.swift
```diff
@@ -0,0 +1,48 @@
...
+    init(id: Int, time: TimeInterval?, lottie: LottieAnimation, values: [String:String] = [:]) {
+        self.id = id
+        self.time = time ?? lottie.duration
+        self.lottie = lottie
+        self.values = values
+    }
+    
+    init(from apiUnit: WrappedAPIUnit, id: Int) async throws {
+        self.id = id
+        guard let anim = await LottieAnimation.loadedFrom(url: apiUnit.lottieUrl) else {
```
JHawk0224, line 31: Why do `WrappedAPIUnit` and `WrappedUnit` need to be different structs? I'm assuming it's to do with loading this animation, and I'm guessing that this `.loadedFrom` should not be done in advance or something, but would you be able to clarify?

E.g. Why can't we just use `WrappedAPIUnit` everywhere, and then just call some function on it that does `LottieAnimation.loadedFrom` when it is first needed or something?

PennMobile/Wrapped/WrappedUnitView.swift
JHawk0224: It's getting the fractional part of the value. It uses the fact that the fractional part of some number is $x-\left\lfloor x\right\rfloor$, and then just plugging in
```swift
x = CGFloat(vm.activeUnitProgress) * (unit.time / unit.lottie.duration) 
```

Couldn't tell you why this is what you want though as I have no idea how lottie's `currentProgress` works. Would you be able to clarify?

PennMobile/Wrapped/WrappedUnitView.swift
```diff
@@ -0,0 +1,47 @@
...
+                    } else if (vm.activeUnit == unit) {
+                        vm.play()
+                    }
+                }
+        }
+            
+    }
+    
+    func getAngle(_ proxy: GeometryProxy) -> Angle {
+        let progress = proxy.frame(in:.global).minX / proxy.size.width
```
JHawk0224, line 41: Does this make sense if the proxy width is not the entire screen? Idk exactly what this is trying to measure since idk what this proxy wraps atm, but something about this formula just feels off to me (since `proxy.size.width` is relative to the proxy while the `.frame(in: .global).minX` is relative to the whole screen).

What is this measuring (visually)?

PennMobile/Wrapped/WrappedExperienceViewModel.swift
```diff
@@ -0,0 +1,48 @@
...
+    
+    func error(error: any Error) {
+        self.error = error
+        self.wrappedExperienceState = .error
+    }
+}
+
+
+public enum WrappedExperienceState {
+    case loading, landing, error, active, inactive, shareScreen, finished
```
JHawk0224, line 47: I think I can infer what each of these are, but can you add a short comment to each one anyway?

PennMobile/Wrapped/WrappedContainerViewModel.swift
```diff
@@ -0,0 +1,127 @@
...
+
+class WrappedContainerViewModel: ObservableObject {
+    
+    let units: [WrappedUnit]
+    var timer: Timer?
+    
+    @Published var activeUnit: WrappedUnit {
+        didSet {
+            activeUnitProgress = 0.0
+            self.nextUnit = units.sorted(by: {$0.id < $1.id}).filter({$0.id > activeUnit.id}).first
```
JHawk0224, line 20: I'm assuming we know these ids are in increasing order normally. Does backend create the ids? If so, are they not just incremental 1 by 1?

Also _extremely_ small nitpick, but it's technically faster to filter first and then sort later.

PennMobile/Wrapped/WrappedContainerViewModel.swift
```diff
@@ -0,0 +1,127 @@
...
+            self.prevUnit = units.sorted(by: {$0.id > $1.id}).filter({$0.id < activeUnit.id}).first
+        }
+    }
+    @Published var nextUnit: WrappedUnit?
+    @Published var prevUnit: WrappedUnit?
+    @Published var activeUnitProgress: CGFloat = 0.0 {
+        didSet {
+            print(activeUnitProgress)
```
JHawk0224, line 28: Should this be removed?

PennMobile/Wrapped/WrappedContainerViewModel.swift
```diff
@@ -0,0 +1,127 @@
...
+                lastFire = nil
+                activeUnitPlaybackMode = .paused(at: .frame(0))
+            case .paused, .finished:
+                timer?.invalidate()
+                timer = nil
+                lastFire = nil
+                activeUnitPlaybackMode = .paused(at: .currentFrame)
+                
+            case .playing:
+                timer = Timer.scheduledTimer(withTimeInterval: 0.0, repeats: true) {_ in
```
JHawk0224, line 52: Is it okay to set this to 0.0 or should there be some minimum framerate? I don't know what the standard is but just wondering.

PennMobile/Wrapped/WrappedContainerView.swift
```diff
@@ -0,0 +1,118 @@
...
+                            ZStack {
+                                Rectangle()
+                                    .size(width: CGFloat((Float(proxy.size.width))), height: 2)
+                                    .cornerRadius(1)
+                                    .foregroundStyle(.white.opacity(0.5))
+                                if unit.id <= containerVM.activeUnit.id {
+                                    Rectangle()
+                                        .size(width: unit.id < containerVM.activeUnit.id ? proxy.size.width :
+                                                CGFloat((Float(proxy.size.width) * Float(containerVM.activeUnitProgress))), height: 2)
+                                        .cornerRadius(1)
```
JHawk0224, line 47: You could also just use `RoundedRectangle()` since `.cornerRadius` is deprecated

PennMobile/Wrapped/WrappedContainerView.swift
```diff
@@ -0,0 +1,118 @@
...
+        .onLongPressGesture(perform: {
+            containerVM.pause()
+            UIImpactFeedbackGenerator(style: .soft).impactOccurred()
+        }, onPressingChanged: { pressed in
+            if !pressed && containerVM.state == .paused {
+                containerVM.play()
+            }
+        })
+        .onTapGesture { location in
+            if (location.x > 150) {
```
JHawk0224, line 79: I'm assuming by default `.onTapGesture` uses the local coordinate space, and that somewhere else there's a frame constraining this view to width 300? Or is this just the entire screen

PennMobile/Wrapped/WrappedContainerView.swift
```diff
@@ -0,0 +1,118 @@
...
+                break;
+            case .active:
+                containerVM.play()
+                break;
+            @unknown default:
+                break;
+            }
+        }
+        .onChange(of: containerVM.state) { new in
+            if (new == .finished) {
```
JHawk0224, line 107: (ew parentheses)

PennMobile/Wrapped/WrappedExperienceViewModel.swift
```diff
@@ -0,0 +1,48 @@
...
+    
+    let model: WrappedModel
+    @Published var wrappedExperienceState: WrappedExperienceState = .inactive {
+        didSet {
+            showWrapped = wrappedExperienceState != .inactive
+        }
+    }
+    @Published var showWrapped: Bool = false
+    @Published var error: Error?
+    @Published var containerVM: WrappedContainerViewModel?
```
JHawk0224, line 21: So the `WrappedExperienceViewModel` view model is really only just to measure the overall state of wrapped? Like if it errored or is showing, but `WrappedContainerViewModel` handles all the play/unit logic?

PennMobile/Wrapped/WrappedLoadingView.swift
```diff
@@ -0,0 +1,53 @@
...
+    var onFinish: (Bool) -> Void
+    
+    @EnvironmentObject var vm: WrappedExperienceViewModel
+    @State var progress: Double = 0
+    
+    func start() async {
+        var units: [WrappedUnit] = []
+        if let pages = vm.model.pages {
+            for page in pages {
+                let val = await LottieAnimation.loadedFrom(url: page.lottieUrl)
```
JHawk0224, line 25: I don't fully understand the data flow, but with regards to my earlier comment about the separation between `WrappedAPIUnit` and `WrappedUnit`, could this `val` just be an optional field in the prior only loaded when needed?

PennMobile/Wrapped/WrappedLoadingView.swift
```diff
@@ -0,0 +1,53 @@
...
+                if let anim = val {
+                    units.append(WrappedUnit(id: page.id, time: page.time, lottie: anim, values: page.values))
+                }
+                
+                progress += 1 / Double(pages.count)
```
JHawk0224, line 30: Btw I haven't really been checking this play/pause/timer logic everywhere (or tracking the states of the view models) as it's pretty hard to follow. I'm thinking that as long as it works I trust it. Do you want me to look into it deeper?

PennMobile/Wrapped/WrappedLoadingView.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+    
+    var body: some View {
+        LottieView {
+          try await DotLottieFile.named("WrappedLoading")
+        }
+        .playing(loopMode: .autoReverse)
+        .frame(width: 250, height: 250)
+        .onAppear {
+            Task {
```
JHawk0224, line 44: `.task {`?

PennMobile/Wrapped/WrappedLoadingView.swift
```diff
@@ -0,0 +1,53 @@
...
+            vm.wrappedUnits = units.sorted(by: { $0.id < $1.id })
+        }
+        onFinish(!units.isEmpty)
+    }
+    
+    var body: some View {
+        LottieView {
+          try await DotLottieFile.named("WrappedLoading")
+        }
+        .playing(loopMode: .autoReverse)
```
JHawk0224, line 41: What does `.autoReverse` do?

JHawk0224 (approved):
I haven't had a chance to test it yet, but if it works then it looks great! Most of my questions are just personal clarifying questions to understand how it works.

Looks great!

PennMobile/Wrapped/WrappedModel.swift
jonathanmelitski: I did the safely-decoded array to ensure that we create an array of valid LottiePages. In hindsight (given that it's coming from a django backend) it _could_ be safe to just use the array decoder, but this functionality allows me to skip over pages that, say, have weird values (i.e. invalid URL or timecode string).

PennMobile/Wrapped/WrappedAPIUnit.swift
jonathanmelitski: Yes

PennMobile/Wrapped/WrappedModel.swift
jonathanmelitski: It would mark the whole array as an invalid decode I think, but per my other [comment](https://github.com/pennlabs/penn-mobile-ios/pull/602#discussion_r2070377166), I want to make sure we have an array of only valid pages. Open to other ways if you think this is either overly bloated, or an excessive amount of safety.

PennMobile/Wrapped/WrappedUnitView.swift
jonathanmelitski: Ok I remember now: will probably comment very similar in my next commit:

- The design spec we agreed on is that if the time given by backend is greater than the duration of the animation, loop for the desired time.
- Lottie's `currentProgress` takes a CGFloat from 0 to 1.0.
- `vm.activeUnitProgress` represents the total duration of the animation (which should be a floating point value (0.0...1.0) consistent with the duration given by backend)
- If I were to assign `.currentProgress` to `vm.activeUnitProgress`, the animation speed would be scaled to the duration given by backend, so
- We calculate what I will call a wrap-around value, which is $\lfloor x \rfloor$ in your comment, which normalizes the progress to a float between 0 and 1 for all animations which may or may not be longer than the duration expected by backend.

Caveats:
- Yes, this means I am manually telling Lottie to change the frame/progress, which is probably a performance hit, instead of using their built-in player.
- **I may have just discovered a bug, so I'm probably going to rewrite/rethink this logic to try to make it easier to understand.**

jonathanmelitski: Ok I just discovered why I'm so confused: the parentheses make it unclear.

`(vm.activeUnitProgress * unit.time) / unit.lottie.duration` is a way more clear way to understand what's going on. We're basically normalizing the progress to the length of the animation. Final explanation:

$$
\frac{activeUnitProgress \times backendTime}{animDuration} - \left \lfloor \frac{activeUnitProgress \times backendTime}{animDuration} \right \rfloor
$$

is a normalized duration for the animation. Consider the cases:

$backendTime = animDuration$: In this case, this whole value will cancel out to $activeUnitProgress$, since it's a value 0f...1.0f

$backendTime < animDuration$: (expected behavior, cut off animation at desired backend time) In this case, we would only want to play some percent of the total animation, so whatever the current progress is should be normalized to some value between 0 and that percent. That percentage value is $\tfrac{backendTime}{animDuration}$. This means that as activeUnitProgress goes from 0 to 1, the animation's progress goes from 0 to some desired progress, and will stop at that progress. (the floor value always cancels in this case)

$backendTime > animDuration$: (expected behavior, loop animation) In this case, we have to handle the idea that just multiplying the active progress by some sum could yield a value > 1 (which causes the animation to break). Say we have an animation that is 3 seconds that we want to play for 9 seconds on repeat. That fraction from before = 3, so to accurately play on a loop, we should play from progress 0...1, then 0...1, then 0...1. This is the same as subtracting the floor value. So the progress would look like $0...1 - 0$ then $1.0...2.0 - 1$ then $2.0...3.0 - 2$, which achieves the desired behavior.

Let me be clear, this is solely because I have to manually control playback (to enable pausing behavior, since Lottie couldn't figure out how to do that apparently). I'm going to abstract this sum into a variable and link to this comment.

PennMobile/Wrapped/WrappedModel.swift
jonathanmelitski: @JHawk0224

PennMobile/Wrapped/WrappedAPIUnit.swift
jonathanmelitski: I tried it out and I agree with the notion of readability (although I'm not disputing how cool that is)

PennMobile/Wrapped/WrappedLoadingView.swift
jonathanmelitski: note that progress was for when I had a loading indicator for just the wrapped loading state. The actual wrapped Unit Progress is a bit hacky, since I had to bypass the built-in player states. See the long comment about calculating the progress for an explanation.

Basically there's a ticking timer that calculates the progress, and such timer is pausable (since changing the state to paused on a long hold gesture wouldn't pause the animation for some reason)

PennMobile/Wrapped/WrappedLoadingView.swift
jonathanmelitski: It probably can be now that the wrapped field from backend has a dedicated ID. will change.

PennMobile/Wrapped/WrappedExperienceViewModel.swift
jonathanmelitski: Yes this precedes this SP25 implementation of wrapped, but ideally we can have a loading state, a finished/share state, and a player state, hence two separate view models? I thought that would make sense in the practical case (and the fact that this project was being brought in from a separate project).

PennMobile/Wrapped/WrappedLoadingView.swift
jonathanmelitski: progress: $0.0 \rightarrow 1.0 \rightarrow 0.0$

PennMobile/Wrapped/WrappedLoadingView.swift
jonathanmelitski: Agreed.

PennMobile/Wrapped/WrappedUnit.swift
jonathanmelitski: Agreed.

jonathanmelitski: I did that when we weren't exactly sure what the model response from backend would be so I wanted to safeunwrap the models. I think it's okay now since it's coming from a stable Django backend.

PennMobile/Wrapped/WrappedModel.swift
jonathanmelitski: Resolved with removal of WrappedAPIUnit

PennMobile/Wrapped/WrappedExperienceViewModel.swift
jonathanmelitski: Yes.

PennMobile/Wrapped/WrappedUnitView.swift
jonathanmelitski: So this proxy measures the percent of the view that is on the screen and is used for the current-obsolete swipe cube gesture. So the 3D cube rotation effect basically takes the percentage of viewport usage as a factor in calculating the rotation effect.

jonathanmelitski: Clarification: depends on the use of TabView and TabViewMode page to allow native swiping, just applies the effect as a function of how much the user swipes the view off the screen.

jonathanmelitski: Again this is currently obsolete, since swipe actions are disabled, but figured it should be left there.

PennMobile/Wrapped/WrappedContainerViewModel.swift
jonathanmelitski: The ID is a user-defined value. My thought is that sorting by id allows for conditional wrapped pages. If we always play in order, we can play 1,2,3 for every user,  4-5 for conditional users, then 6-etc. for everyone else, and it will always be in order, without having to manage the IDs for many users.

Considering now that there should be a write-up of this in Notion or something. There were a lot of design decisions (like this one) that went into this that are not going to be apparent in the future.

PennMobile/Wrapped/WrappedContainerViewModel.swift
jonathanmelitski: Yes.

PennMobile/Wrapped/WrappedContainerView.swift
jonathanmelitski: java dev at heart

PennMobile/Wrapped/WrappedContainerViewModel.swift
jonathanmelitski: I can consider the framerate of the animation, which is in the model. I don't think it matters but is probably better for system performance.

PennMobile/Wrapped/WrappedContainerView.swift
jonathanmelitski: It's just the entire screen lol

jonathanmelitski: my thought is that for such a specific implementation (where we know wrapped is being displayed via fullScreenCover), this is probably okay, but open to suggestions.

jonathanmelitski: https://developer.apple.com/documentation/swiftui/view/ontapgesture(count:coordinatespace:perform:)-21n4i

PennMobile/Wrapped/WrappedLoadingView.swift
jonathanmelitski: Up to you if you want to look at it.

PR #606, opened 2025-05-09
Fix GSR State Management

jonathanmelitski:
The GSR rooms array does not reset on a failed request. This leads to us sending a request for a room at a different location than it belongs. This is not validated by anybody, meaning the request is allowed to go through. kudos @ryantanen 

**Before**
https://github.com/user-attachments/assets/dda7bc46-72b0-4583-8a1b-b1c425298ed8

**After**
https://github.com/user-attachments/assets/cb597909-2276-4c0a-afd5-80933c793533

This was done by setting the rooms array to empty when the request fails, then updating the view accordingly.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -11,7 +11,7 @@
 #
 
 # Uncomment the line if you want fastlane to automatically update itself
-update_fastlane
+# update_fastlane
 
 default_platform(:ios)
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,3 @@
 source "https://rubygems.org"
 
-gem 'fastlane', '~> 2.227', '>= 2.227.1'
+gem 'fastlane', '~> 2.227'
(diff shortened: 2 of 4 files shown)


jonathanmelitski:
Ok so I guess I broke CI, so I'm going to fix that

fastlane/Fastfile
```diff
@@ -11,7 +11,7 @@
 #
 
 # Uncomment the line if you want fastlane to automatically update itself
-update_fastlane
+# update_fastlane
```
JHawk0224, line 14: Do we know why this was breaking?

JHawk0224 (approved):
LGTM!

fastlane/Fastfile
jonathanmelitski: Yeah, for some reason a new version of ruby started updating fast lane, then erroring when it couldn't find the old version.

It always autoupdated but it wasn't a problem until I upgraded ruby bundler version.

Figured this was a fix that makes sense so it doesn't spontaneously break in the future.

PR #609, opened 2025-08-31
Performance and usability improvements to redesigned GSR booking flow

anli5005:
This PR:

* Makes the entire region of the GSR location row view tappable
* Uses `LazyVStack` views for the GSR grid because using it almost crashed my Vision Pro
* Uses the `.scrollView` coordinate space to position the time header, to avoid layout mismatches when swiping to go back


jonathanmelitski (approved):
lazy lazy lazy lazy

LGTM

jonathanmelitski:
The goal was to crash your vision pro btw.

LGTM

PR #612, opened 2025-09-04
Revamp localized descriptions for NetworkingError

anli5005:
Previously, any thrown `NetworkingError`s were displayed with a generic error message:

> The operation couldn't be completed. (PennMobileShared.NetworkingError error 2.).

This PR fixes that by declaring conformance to `LocalizedError`, and adds some more user-friendly error messages to boot.


anli5005:
thank you copilot

anli5005:
never mind

(for context, turns out those strings actually come from the server and we initialize our errors using those strings as raw values in a few places)

PR #613, opened 2025-09-04
Remove the subletting banner

anli5005:
People are probably not looking for housing, so it makes sense to remove this for now to prioritize Portal posts.

In the future, we can consider reinstating this, either with some additional logic or by re-implementing it as a Portal post (although this may change with Marketplace coming soon.)

### DIFF
--- a/PennMobile/Home/HomeViewModel.swift
+++ b/PennMobile/Home/HomeViewModel.swift
@@ -40,7 +40,7 @@ extension Optional {
     }
     
     func populateSublettingBannerData() {
-        data.showSublettingBanner = !UserDefaults.standard.bool(forKey: Self.sublettingBannerKey)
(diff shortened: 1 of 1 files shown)


PennMobile/Home/HomeViewModel.swift
```diff
@@ -40,7 +40,7 @@ extension Optional {
     }
     
     func populateSublettingBannerData() {
-        data.showSublettingBanner = !UserDefaults.standard.bool(forKey: Self.sublettingBannerKey)
+        // data.showSublettingBanner = !UserDefaults.standard.bool(forKey: Self.sublettingBannerKey)
```
julianweng, line 43: How is UserDefaults data managed? If the banner is re-added what would this data look like for existing vs new vs joined (during interim) users?

jonathanmelitski (approved):
We love 1-liner PRs.

LGTM, I agree that we should have something ready (or should otherwise be thinking about) for when we come out with marketplace.

PR #615, opened 2025-09-07
Tweak CI to reduce usage and allow downloading of artifacts

anli5005:
This makes a number of tweaks to our CI/CD workflow, in particular:
* Limits our workflow to run only on pushes to `main`, on PR updates, or via manual dispatch
* Removes the `cleanup` step so we can download artifacts after the fact
* Adds a 14-day retention period for artifacts
* Changes the build numbering scheme to a date-based one (YYYYMMDD.HHMMSS)

The intended effect is two-fold:
1. It will reduce credit usage by not running CI for experimental branches or branches without a PR
2. It will allow us to release hotfixes by downloading the artifact and uploading the .ipa manually to App Store Connect
3. It will prevent most build number conflicts by using a stateless approach

(Storing artifacts will result in increased storage costs, though this should be offset by limiting the range of possible triggers.)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,6 +2,9 @@ name: CI
 
 on:
   push:
+    branches:
+      - 'main'
+  pull_request:
   workflow_dispatch:
 jobs:
   build:
@@ -85,6 +88,7 @@ jobs:
         with:
           name: build
           path: PennMobile.ipa
+          retention-days: 14
   upload:
     needs: build
     if: github.ref == 'refs/heads/main'
@@ -111,12 +115,3 @@ jobs:
           name: build
       - name: Upload to TestFlight
         run: bundle exec fastlane ci_pilot
-  cleanup:
-    runs-on: macos-15
-    needs: [build, upload]
-    if: always()
-    steps:
-      - name: Delete artifacts
-        uses: geekyeggo/delete-artifact@v5
-        with:
-          name: build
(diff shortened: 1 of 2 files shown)


anli5005:
we love vibe coding

.github/workflows/main.yml
```diff
@@ -110,13 +114,4 @@ jobs:
         with:
           name: build
       - name: Upload to TestFlight
-        run: bundle exec fastlane ci_pilot
-  cleanup:
```
jonathanmelitski, line 114: My concern with retaining artifacts after the build is that it removes a point of friction for people reverse engineering our API keys. If people can just go onto our Github and download a full build, it's way easier to reverse engineer.

fastlane/Fastfile
```diff
@@ -30,9 +30,18 @@ platform :ios do
...
+    previous_tz = ENV['TZ']
+    begin
+      ENV['TZ'] = 'America/New_York'
+      date_build_number = Time.now.strftime("%Y%m%d.%H%M%S")
+    ensure
+      ENV['TZ'] = previous_tz
+    end
+
+    increment_build_number(
+      build_number: date_build_number
```
jonathanmelitski, line 43: did you confirm that app store connect accepts non numberic build numbers?

.github/workflows/main.yml
anli5005: We *could* encrypt or upload the binary elsewhere, but tbh I'm not terribly convinced this is much of a concern.

jonathanmelitski (approved):
LGTM

Would be good to have further clarification on why exactly our CI doesn't fully support this. Can't we manually associate builds with certain versions?

PR #625, opened 2025-09-27
HomeViewAnnounceable

jonathanmelitski:
Make it so ViewModels can conform to a protocol called "HomeViewAnnounceable", where they can provide announcements that should appear on the homescreen.


PennMobile/Home/HomeViewModel.swift
```diff
@@ -24,6 +24,8 @@ extension Optional {
 @MainActor class StandardHomeViewModel: HomeViewModel {
```
anli5005, line 24: tbh I think a general refactor of `HomeViewModel` is overdue, but we can punt that to later

anli5005 (approved):
I'm going to approve this for now so as not to block @0kutsu's newbie project, but we should address the comments once his project is done

PR #626, opened 2025-09-28
Switch from groups to folders

anli5005:
With newbies incoming, I thought now would be a good time to take a second crack at this. Hopefully we can eliminate project.pbxproj merge conflicts once and for all.

If CI cooperates, that is.


jonathanmelitski:
@anli5005 but I like dealing with those merge conflicts they force me to be confident about the files I'm creating.

PennMobile.xcodeproj/project.pbxproj
jonathanmelitski, line 1: wait so you actually nuked pbxproj

anli5005: yep it's so freeing

NotificationServiceExtension/NotificationService+ImageCacheing.swift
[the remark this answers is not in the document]
anli5005: well it's in the Kingfisher podspec:

https://github.com/onevcat/Kingfisher/blob/43cc4a1ad0b85b0c79d5ab3987c66c4f7a395aa4/Kingfisher.podspec#L5

PR #627, opened 2025-10-03
Add a feature flagging system

anli5005:
Adds a system to define and configure feature flags so we can merge stuff in without worrying about breaking prod as much.

I'll probably write actual docs about this later, but pretty much:
* Feature flags are added in FeatureFlags.swift using the `@FeatureFlagDefinition` property wrapper
* Feature flags can either be on for everyone (App Store), on for people in the TestFlight, on for running from Xcode, or always off unless manually overriden (Ad-Hoc).
* Once defined, feature flags can be checked simply by doing something like `if FeatureFlags.shared.blahblahblah`
* TestFlight and dev users can manually tweak feature flags to enable/disable experiments. Feature flags can also be forced on/off via environment variables by configuring an Xcode scheme. App Store users will not see the feature flag menu at all, unless their settings were carried over from a TestFlight version.

Here's what the feature flag settings look like:
Feature flag settings


PennMobile/Home/HomeView.swift
```diff
@@ -52,6 +53,11 @@ struct HomeView<Model: HomeViewModel>: View {
                         .padding(.bottom)
                         .multilineTextAlignment(.center)
                         
+                        if FeatureFlags.shared.testFeatureFlag {
```
jonathanmelitski, line 56: this is meant to be left in prod, right?

PennMobile/More Tab/FeatureFlagSettingsView.swift
```diff
@@ -0,0 +1,136 @@
...
+            UserDefaults.group.set(channelOverride?.rawValue, forKey: RolloutChannel.overrideDefaultsKey)
+        }
+    }
+    
+    var currentChannel: RolloutChannel {
+        channelOverride ?? RolloutChannel.inferred
+    }
+    
+    @State var flagOverrides = Dictionary(uniqueKeysWithValues: FeatureFlags.shared.configurableFlags.compactMap { flag in
+        if UserDefaults.group.object(forKey: flag.defaultsKey) != nil {
```
jonathanmelitski, line 24: `if let`?

PennMobileShared/Feature Flags/FeatureFlags.swift
```diff
@@ -0,0 +1,192 @@
...
+    }
+}
+
+public enum RolloutChannel: Int, Comparable, CaseIterable, Hashable {
+    case appStore = 1
+    case testFlight = 2
+    case experimental = 3
+    case adHoc = 4
+    
+    public static let configurable = [RolloutChannel.appStore, .testFlight, .experimental]
```
jonathanmelitski, line 42: like I have so many thing to say about this

PennMobile/More Tab/MoreView.swift
```diff
@@ -87,6 +88,16 @@ struct MoreView: View {
                 Text("Account")
             }
             
+            if FeatureFlags.shared.showFeatureFlagSettings {
```
jonathanmelitski, line 91: Maybe this is a bit of a nit, but I feel like it's a bit prone to issues if the menu that reveals the feature flags and settings is contingent on a feature flag?

Like I feel like it will never go wrong but still feels weird.

PennMobileShared/General/Extensions.swift
```diff
@@ -780,3 +780,7 @@ public extension String {
         return formatter.string(from: NSNumber(value: value)) ?? ""
     }
 }
+
+public extension UserDefaults {
+    static let group = UserDefaults(suiteName: Storage.appGroupID)!
```
jonathanmelitski, line 785: not sure I'm a fan of a new user defaults suite for this feature specifically. Like this could lead to insane unexpected behavior in the future.

jonathanmelitski (changes requested):
Gave it a once-over and I'll probably have more comments, but here's this for now.

PennMobile/Home/HomeView.swift
anli5005: Yes

PennMobile/More Tab/MoreView.swift
anli5005: `showFeatureFlagSettings` isn't actually a computed property, and while I agree it's a bit weird, there's a decent bit of logic that goes on in determining this beyond the feature flag:

* If running on TestFlight or a similar environment, the settings are always shown
* If we see that you've overridden at least one feature flag (with the exception of `ALWAYS_SHOW_FEATURE_FLAG_SETTINGS`), we show the settings
* If the `ALWAYS_SHOW_FEATURE_FLAG_SETTINGS` feature flag is set (which can be useful for people who switch between the TestFlight and App Store versions), we show the settings
* Otherwise, we don't show the settings

PennMobileShared/General/Extensions.swift
anli5005: We do this so that feature flags are shared between all of our app's extensions. The app group ID isn't specific to feature flagging, it's just the first time we're using `UserDefaults` in the app group. (We already do this for some `Storage` classes.)

jonathanmelitski: we should just rewrite penn mobile at this point

jonathanmelitski (approved):
lgtm (lead doesn't trust me)

PR #630, opened 2025-10-06
Remove force unwrapping of currentMenu Binding

anli5005:
Should fix a recurring crash we've been seeing. Affected crash reports usually display something like this:

> SwiftUICore: BindingOperations.ForceUnwrapping.get(base:) + 308

I'm guessing this is because projected Bindings probably aren't perfectly synced with their underlying values? Either way, I suppose this is your daily reminder to not force unwrap things unless absolutely necessary lol

### DIFF
--- a/PennMobile/Dining/SwiftUI/Views/Venue/Detail View/DiningVenueDetailMenuView.swift
+++ b/PennMobile/Dining/SwiftUI/Views/Venue/Detail View/DiningVenueDetailMenuView.swift
@@ -79,9 +79,10 @@ struct DiningVenueDetailMenuView: View {
         LazyVStack(pinnedViews: [.sectionHeaders]) {
             HStack {
                 if currentMenu != nil {
-                    Picker("Menu", selection: Binding($currentMenu)!) {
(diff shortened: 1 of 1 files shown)


PennMobile/Dining/SwiftUI/Views/Venue/Detail View/DiningVenueDetailMenuView.swift
```diff
@@ -79,9 +79,10 @@ struct DiningVenueDetailMenuView: View {
         LazyVStack(pinnedViews: [.sectionHeaders]) {
             HStack {
                 if currentMenu != nil {
-                    Picker("Menu", selection: Binding($currentMenu)!) {
+                    Picker("Menu", selection: $currentMenu) {
```
jonathanmelitski, line 82: @my_newbie_project

anli5005: tbf this isn't an entirely unreasonable thing to do

PR #632, opened 2025-10-06
Add Font Provider Before Backend

jonathanmelitski:
**NOTE: THIS SHOULD NOT BE MERGED BEFORE BACKEND FINISHES THEIR IMPLEMENTATION _AND_ IT HAS BEEN ADEQUATELY TESTED** (_but it should work_)

Dynamically rendered text requires the original font used for the design (instead of non-dynamic text, which relies on vector graphics). For 2025S, we just [hardcoded the fonts](https://github.com/pennlabs/penn-mobile-ios/pull/603) but alluded to an AnimationFontProvider-conforming class, which is this PR.

Backend should provide a dictionary of `[String: URL]` whose keys are individual font names, and whose URLs are links to ttf font files. This way, we download all the necessary fonts for wrapped in the loading phase, then allow the animation to reference these fonts. Again, this hasn't been tested, since there's no backend infra for this. However, the changes required as a result of their implementation will be minimal.

**Other things:**
- Using `api/wrapped/current` instead of a hard-coded semester. This is the real reason that this PR should wait, since they currently don't error handle unknown semester names, so it was causing 2.5k sentry errors. But this is in the works (where they can identify a certain semester to be pointed to from /current)
- Concurrent loading courtesty of TaskGroup


jonathanmelitski:
> "feature flag this as ad-hoc and merge it in before backend is ready"

I don't want it on prod devices until at least /current is handled, because it will just cause a spike in sentry errors.

PennMobile/Wrapped/WrappedModel.swift
```diff
@@ -47,5 +64,49 @@ public struct WrappedModel: Decodable {
         }
         
         self.pages = newPages.filter({ $0.lottie != nil }).sorted(by: { $0.id < $1.id })
```
jonathanmelitski, line 66: @anli5005 This isn't what you mean but WrappedUnitView cannot display before we have a loaded model

PennMobile/Home/HomeViewModel.swift
```diff
@@ -196,30 +196,25 @@ extension Optional {
...
-                    DispatchQueue.main.async {
-                        do {
-                            let wrapped = try JSONDecoder().decode(WrappedModel.self, from: data)
-                            self.data.wrapped = .success(wrapped)
-                        } catch {
-                            self.data.wrapped = .failure(error)
-                        }
-                    }
+            
+            DispatchQueue.main.async {
```
anli5005, line 210: `await MainActor.run`

Or just mark the entire `Task` as `@MainActor`

PR #662, opened 2026-03-21
added feedback button to bottom of home page

wyq54123-ux:


khoiddinh (approved):
Looks good to me. In the future, please name your branch with an identifier.

i.e. instead of "feedback-button-on-main" if I was building the feature it would be "khoidinh/feedback-button-on-main." Feel free to pick an identifier yourself it doesn't matter what it is but helps to keep track of who is doing what.

PR #673, opened 2026-04-05
Fix Auth Stability Issues & bump LabsPlatformSwift version

jonathanmelitski:
With https://github.com/pennlabs/labs-platform-swift/pull/6 a bump is required.

Other changes (some, not all, required from the aforementioned bump):

- Supporting a synchronous initial login state check to allow login logic to take place before anything else.
- Initialize AuthManager state prior to main thread establishment to prevent weird UX with the login screen flashing first (when a refresh flow is required). Required a change to `determineInitialState` (see below)
- Force-fetching the home view model on login change (caused the home page to load correctly on first login without requiring a manual refresh)
- `determineInitialState` has been modified to instead make state changes w.r.t. auth state more explicit. Now called `getInitialState`, where changes to AuthManager.authState are explicitly declared. (immutable functions >>>>)


khoiddinh (approved):
LGTM

PR #675, opened 2026-04-20
fixed home view announceable race condition

khoiddinh:
There was a bug in fetch that it was building the list of announceable by appending to a list, so if it was called twice it was possible to have duplicate "new feature entries in home view". Fixed by building in async task then updating list when done to make it quasi-atomic.


jonathanmelitski (approved):
A task group would be a better approach since all the requests could take place in parallel, but I'm not going to block on it.

PR #676, opened 2026-04-22
gsr share UI simplifying

wyq54123-ux:
Compressed GSR reservations detail pages into individual cells on the Reservation tab.
Also simplified the share preview.


PennMobile/GSR-Booking/Views/Reservations/GSRShareSubViews/GSRReservationContentView.swift
```diff
@@ -19,17 +19,29 @@ struct GSRReservationContentView: View {
     
     var body: some View {
         let gsrLocation = model.gsr.name
-
-        ScrollView {
+        
+        ScrollView() {
```
khoiddinh: no need to have ()

PennMobile/GSR-Booking/Views/Reservations/GSRShareSubViews/GSRReservationContentView.swift
```diff
@@ -55,12 +67,6 @@ struct GSRReservationContentView: View {
 
                     Divider()
 
-                    GSRMapSectionView(
-                        model: model,
-                        gsrLocation: gsrLocation,
-                        roomName: roomName
-                    )
-
                     if !mode.isReadOnly {
```
khoiddinh: I don't think we need .isReadOnly, this is a legacy of when we had a reservationdetailview for people that own gsr reservations, but now since we removed that with the rewrite we can remove this, gsrdetailview should only be read only

khoiddinh (changes requested):
There's a lot of code that is now obsolete (i.e. GSRMapView, DeleteButtonView). Also see comments about not needed the GSR mode variable since detailview is now only read only. Another (minor) comment is that the toast when you add it to iCal is red which is kind of weird if it's like (Added to Calendar!). Maybe adding different colors to the toasts would be nice, but also not necessarily the point of this feature. We could push that off to a later date.

khoiddinh (approved):

PR #679, opened 2026-09-02
My attempt at fixing the dining graph (part 1 of `undefined`)

jonathanmelitski:
Fix the graph because I'm a good developer. 

Pull out a lot of the logic that sought to clean the data. Data is assumed to be clean from the source, and if this assumption proves to be false, we will adjust accordingly.

Also migrated to a simple linear regression starting at the latest increase in value (in the event that people buy more swipes or something).

Still need to recover auth state more gracefully (but this will likely come with a larger change to the main auth flow).


PennMobileShared/General/Extensions.swift
jonathanmelitski, line 1: co-director btw @khoiddinh @anli5005

khoiddinh: oh god

jonathanmelitski: also why this logic hasn't been moved to mobile-backend like 3 years ago is a mystery to me

khoiddinh:
You should remove the SwiftUI folder its useless just move contents up to the parent.

khoiddinh:
In a similar vein, refactor the file names why is SwiftUI in there

khoiddinh (approved):
LGTM

PR #682, opened 2026-09-05
Move Dining to WebAuthenticationView

jonathanmelitski:
With LabsPlatformSwift moving to Apple's inbuilt WebAuthenticationSession, moving the rest of the authentication services (namely Dining Analytics) to the same flow was sensible.

This PR:
1. Extracts the core OAuth logic out of a web view delegate
2. Removes the "Would you like to log into campus express" alert, since WebAuthenticationSession already prompts.
3. Refactors the logic using a single async function `login` which is called when the dining login is set to presented.

Unfortunately one of the mandates of using the SwiftUI-native WebAuthenticationSession is that you have to extract the session from SwiftUI environment to use it, which prompted this wrapper approach. Refactoring away from that would have been likely a hackier solution, though I'm not super happy about having a wrapper for this.


PennMobile/Dining/Login/DiningLogin.swift
```diff
@@ -0,0 +1,102 @@
...
+class DiningLogin {
+    static let clientId = "5c09c08b240a56d22f06b46789d0528a"
+    static let authorizeUrl = URL(string: "https://prod.campusexpress.upenn.edu/api/v1/oauth/authorize")!
+    static let tokenUrl = URL(string: "https://prod.campusexpress.upenn.edu/api/v1/oauth/token")!
+
+    static let redirectHost = "pennlabs.org"
+    static let redirectPath = "/pennmobile/ios/campus_express_callback/"
+    static var redirectUri: String { "https://\(redirectHost)\(redirectPath)" }
+
+    enum Error: Swift.Error {
```
khoiddinh: can we move away from this, I think we should centralize our error enums

khoiddinh: not gonna block on this but def something to think about when refactoring, about how to standardize these

khoiddinh (approved):
see comments, but wont block, looks good. BTW I think the wrapper approach is very clean

PR #683, opened 2026-09-06
Fix dining analytics edge case

wyq54123-ux:
fixed case when user hasn't spent any dollars/swipes, 
caused because of not checking the case where most recent balance is equal to the most recent max balance


khoiddinh (approved):
You are all fired.

PR #688, opened 2026-09-12
Massive Initial Refactor

jonathanmelitski:
Created a common endpoint model for specific backend calls to be hard-coded prior to frontend needing to manage them.

Still need to figure out the ways in which this should be abstracted. The first main one is a `@BackendFetched` property wrapper which can fetch the values automatically.

Also still need to figure out how caching should take place for a potential offline mode, as well as guest mode, which isn't implemented.

Guest mode will require a minimal state holder that is separate from platform auth, so we know whether or not to try to execute backend endpoints.


PennMobile/Applications/GSR/GSRViewModel.swift
```diff
@@ -0,0 +1,306 @@
...
+    
+    func addToCalendar(_ reservation: GSRReservation) async {
+        do {
+            try await CalendarHelper.addToCalendar(
+                title: "GSR Booking: \(reservation.gsr.name) \(reservation.roomName)",
+                location: reservation.gsr.name,
+                start: reservation.start,
+                end: reservation.end
+            )
+            message = "Added to calendar!"
```
khoiddinh, line 225: do catch should be in the backend/shared

PennMobile/Applications/GSR/GSRViewModel.swift
```diff
@@ -0,0 +1,306 @@
...
+    }
+    
+    private func relevantAvailability(for room: GSRRoom) -> [GSRTimeSlot] {
+        room.availability.filter { Calendar.current.isDate($0.startTime, inSameDayAs: selectedDate) }
+    }
+    
+    private nonisolated static func validate(_ slots: [Selection]) throws {
+        guard let first = slots.first else { return }
+        
+        guard slots.allSatisfy({ $0.room.id == first.room.id }) else {
```
khoiddinh, line 251: we should come up with better ways to standardize frontend errors

khoiddinh: thinking enforce a style thing with a standard error file and one for each subapp

PennMobile/Applications/GSR/GSRViewModel.swift
```diff
@@ -0,0 +1,306 @@
...
+            }
+            try Self.validate(selectedTimeslots)
+            
+            let sorted = selectedTimeslots.sorted { $0.slot.startTime < $1.slot.startTime }
+            let first = sorted.first!
+            let booking = GSRBooking(gid: location.gid, startTime: first.slot.startTime, endTime: sorted.last!.slot.endTime, id: first.room.id, roomName: first.room.roomName)
+            
+            try await PennMobileApplication.GSR.MakeBooking(booking: booking).execute(
+            recentBooking = booking
+        } catch {
```
khoiddinh, line 200: i hate this error pattern

khoiddinh (approved):

PR #690, opened 2026-09-14
fixed dining hall caching issues

khoiddinh:
title


PennMobile/Dining/Dining Halls/DiningViewModel.swift
```diff
@@ -39,47 +39,35 @@ class DiningViewModel: ObservableObject {
...
+            if case .failure(let error) = diningResult {
                 self.alertType = error
             }
-            
-            self.diningVenuesIsLoading = false
+            return
+        }
+
+        let cachedIDs = (try? Storage.retrieveThrowing(DiningVenue.favoritesDirectory, from: .caches, as: [Int].self)) ?? []
+        showVenues(diningVenues, favoritesIDs: cachedIDs)
```
jonathanmelitski: so personally I'm not a fan of outsourcing state changes to other functions. As far as readability is concerned. I would prefer this function return some variables and we edit our state along with our other changes.

I think it would be different if we called this function elsewhere in our code, but it's more difficult to track state changes this way.

PennMobile/Dining/Dining Halls/DiningViewModel.swift
```diff
@@ -39,47 +39,35 @@ class DiningViewModel: ObservableObject {
...
+        guard case .success(let diningVenues) = diningResult else {
+            if case .failure(let error) = diningResult {
                 self.alertType = error
             }
-            
-            self.diningVenuesIsLoading = false
+            return
+        }
+
+        let cachedIDs = (try? Storage.retrieveThrowing(DiningVenue.favoritesDirectory, from: .caches, as: [Int].self)) ?? []
```
jonathanmelitski: in theory favorites are also stored on network as well, so we wouldn't want to pull from cache here under the same logic

PennMobile/Dining/Dining Halls/DiningViewModel.swift
```diff
@@ -39,47 +39,35 @@ class DiningViewModel: ObservableObject {
...
+        showVenues(diningVenues, favoritesIDs: cachedIDs)
+
+        if case .success(let favorites) = await UserDBManager.shared.fetchDiningPreferences() {
```
jonathanmelitski: not a fan of calling our state changes twice here (it just makes it twice as difficult to debug.

My preference would be:
1. `fetchDiningPreferences()` to get the latest favorites from backend.
2. If that fails (and you're still interested in fetching from cache) try to get favorites from cache
3. run your state changes to update the venues accordingly in-line as opposed to this function.

PennMobile/Dining/Dining Halls/DiningViewModel.swift
```diff
@@ -39,47 +39,35 @@ class DiningViewModel: ObservableObject {
   // MARK: - Venue Methods
     let ordering: [VenueType] = [.dining, .retail]
 
+    private func showVenues(_ venues: [DiningVenue], favoritesIDs: [Int]) {
```
jonathanmelitski: not a fan of this, see comments below

jonathanmelitski (changes requested):
This could be consolidated to make it so much more readable and easier to debug.

PennMobile/Dining/Dining Halls/DiningViewModel.swift
```diff
@@ -40,47 +40,33 @@ class DiningViewModel: ObservableObject {
...
+        }
+        
+        
+        var favorites: [Int] = []
+        if case .success(let networkFavorites) = await UserDBManager.shared.fetchDiningPreferences() {
+            favorites = networkFavorites.map(\.id)
+            Storage.store(favorites, to: .caches, as: DiningVenue.favoritesDirectory)
+        } else if let cached = try? Storage.retrieveThrowing(DiningVenue.favoritesDirectory, from: .caches, as: [Int].self) {
+            favorites = cached
+        }
```
khoiddinh, line 64: If we network and cache fails shouldn't it display an error instead of just displaying empty

khoiddinh: like this is the entire issue lol that it can't fail by being empty

